Issue #10, opened 2018-10-29
Add detection/deblending/measurement algorithm


sowmyakth:
Make a function, `btk.measure`,  that inputs `draw_blends` generator, draws blended images and then runs a measurement algorithm on it to output results. These results would then be analyzed using `btk.metrics`  to gauge the performance of the algorithm. The measurement algorithm will be a user input function.  

For reference, a function to perform measurement with the lsst science pipeline can be described in `utils.py`

sowmyakth:
Framework Implemented in [#10](https://github.com/LSSTDESC/BlendingToolKit/pull/10). Framework is general enough for user to input any detection/debelnding/segmentation/measurement algorithm.

PR #11, opened 2018-10-30
#10

sowmyakth:
`measure.py` can be used to run any deblending/ measurement algorithm on the output images from `draw_blends.py`.  An example of using this with DM stack is shown in `notebooks/run_basic.ipynb` with the detection functions defined in `utils.py`


xuod:
Great work @sowmyakth ! This is a great first version of the whole pipeline. Also, one unimportant thing (not worth delaying the pull request): in `utils.py`, lines 24-25, shouldn't the index be 3 for band i?

sowmyakth:
@xuod Yup. you're right, my bad. Will change it in the next PR.

Issue #5, opened 2018-07-27
Implement observing conditions generator


roucelle:
create_observing_generator

- input : survey name (Possibility for the user to give a table of properties ?)
- output : generator

Creates a generator of characteristics of a given observation for a given survey.

generator outputs exposure time, PSF (fixed for now ?), filters, rotation of stamp. 

Could be adapted from WeakLensingDeblending.

sowmyakth:
Implemented in [#14](https://github.com/LSSTDESC/BlendingToolKit/pull/14)

PR #14, opened 2019-01-26
#5 Implement observing conditions generator

sowmyakth:
Different blends in a batch can now have different observing conditions. A user defined function can set the observing condition. The default settings are same as those in WLD. An example of using this is shown in `custom_sampling_function.ipynb`


sowmyakth:
Implemented in [#14](https://github.com/LSSTDESC/BlendingToolKit/pull/14)

Issue #16, opened 2019-01-28
Implement friends-of-friends algorithm to generate blends


sowmyakth:
Extension of #4 to implement friends-of-friend algorithm in creating blend_list in `create_blend_generator`. Currently realistic blends can only be created by identifying groups in a pre-run WLD catalog as shown in notebooks/custom_sampling_function.ipynb shows. A fast on the fly identification of neighbors in a group would eliminate to run the WLD catalog beforehand.

ismael2395:
Hi @sowmyakth , I like this idea, could you elaborate a bit on what it involves or where you thinking of incorporating this FOF algorithm for the blends?

sowmyakth:
@ismael2395 I was thinking an extension of example 4 in this [notebook](https://github.com/LSSTDESC/BlendingToolKit/blob/9e4f0d5fa175592ebd646e433fbc87bcdf9a0d46/notebooks/custom_sampling_function.ipynb). Here galaxies within a certain threshold radius of a central galaxy are classified as part of the blend. It would be nice to also recursively include galaxies within the threshold radius of all those selected galaxies to also be included in the blend. At least that's the idea. @cosmicshear who has implemented this in his unrecognized blends study may have additional input.

ismael2395:
See also this notebook: https://github.com/LSSTDESC/DC2-analysis/blob/master/tutorials/matching_fof.ipynb

enourbakhsh:
@cosmicshear here (renamed my GitHub recently). In case you are interested in using FoF, I made a parallel python package similar in use to `FoFCatalogMatching` which runs >1 order of magnitude faster. Planning on uploading it to PyPI after adding some docstrings.

fjaviersanchez:
@ismael2395, I leave this info here for your consideration too. Please feel free to disregard if it's not useful. I found that `sklearn.neighbors.KDTree` works much faster than `FoFCatalogMatching` if you just want to query neighbors within a certain radius: https://scikit-learn.org/stable/modules/generated/sklearn.neighbors.KDTree.html (I would build a tree with the centroids of the objects themselves and just `query_radius`with the `counts_only = True` option, and check if the returned number is larger than 1 to focus on those, for the radius you can play a bit since it will depend on the FWHM that you use for the images).

enourbakhsh:
[...]

Issue #4, opened 2018-07-27
Implement create_blend_generator


roucelle:
- input : catalog, batch size, sampling strategy 
- output : generator.

The generator creates blend lists according to a given strategy. Proposed strategies include simple 
1. random choice and positioning on the post stamps of a given number of galaxies
2. using friends-of-friends implemented in DC2 to find close pairs.

sowmyakth:
Implemented in [#8](https://github.com/LSSTDESC/BlendingToolKit/pull/8)

PR #19, opened 2019-04-04
#16

sowmyakth:
Tutorial now includes a custom function that picks a galaxy at random from catsim catalog and draws all objects within a fixed radius of it. This is not yet a complete implementation of #4.  

Apart from this, the code is made 2.7 compatible and the documentation in improved.


sowmyakth:
Tutorials now include example of custom functions where the location of galaxies can be randomly set, groups of galaxies can be set as blend, galaxies within a fixed radius are considered a blend.  This is not yet a complete implementation of friend-of-friend algorithm #4.
Script is compatible with python 3.6>=. `draw_blends.py` contains functions non-compatible with python2

PR #30, opened 2020-05-05
Ismael code review

ismael2395:
Hi @sowmyakth , 

I think I finally finished my code-review. It all looks pretty good and really like your organization!
My comments are marked with `# REVIEW`, and are hopefully (?) easy to find and actually review (please let me know if this isn't the case and we can try some other format). 

Most of what I did was doing some (automatic) reformatting of the code to make it consistent, and suggested some small changes to make where things are in the code a bit more explicit/clearer. Two small things: 
- The tests are currently failing and that is just because I defined some abstract classes with abstract methods, but I'm not completely sure if you recommend keeping these methods ( I was confused why they are two methods for `Measurement_params` but they seem to return similar things?) 
- You will also see I have bigger block `# REVIEW` comments, this are mostly just ideas that I had for enhancement. Feel free to chime in and it would be great to discuss what you think, but it's also totally ok if you don't. 

Thanks in advance! 

P.S. I was of course planning to remove all the review comments before merging this PR into master.

### DIFF
--- a/btk/compute_metrics.py
+++ b/btk/compute_metrics.py
@@ -314,8 +314,6 @@ def evaluate_detection(true_tables, detected_tables,
     return batch_true_table, batch_detected_table, batch_blend_summary
 
 
-# REVIEW:
-#  Not implemented yet?
 def evaluate_segmentation(segmentation, data=None, index=None):
     if segmentation is None:
         return None
--- a/btk/compute_metrics.py
+++ b/btk/compute_metrics.py
@@ -1,9 +1,13 @@
-import numpy as np
+from abc import ABC, abstractmethod
+
 import astropy.table
+import numpy as np
 import scipy.spatial
 
 
-class Metrics_params(object):
+# REVIEW:
+#  Also make it an abstract class.
+class Metrics_params(ABC):
     def __init__(self, meas_generator, sim_param):
         """Class describing functions to return results of
          detection/deblending/measurement algorithm in meas_generator. Each
@@ -12,6 +16,7 @@ class Metrics_params(object):
         self.meas_generator = meas_generator
         self.sim_param = sim_param
 
+    @abstractmethod
     def get_detections(self):
         """
         Returns detection results as two catalogs one with entries of true
@@ -100,7 +105,7 @@ def get_m_z_diff(true_table, detected_true):
     dx = true_table['dx'] - match['dx']
     dy = true_table['dy'] - match['dy']
     true_table['ddist_match'] = np.hypot(dx, dy)
-    true_table['dnorm_dist_match'] = np.hypot(dx, dy)/match['size']
+    true_table['dnorm_dist_match'] = np.hypot(dx, dy) / match['size']
 
 
 def initialize_detection_tables(detected_table, true_table,
@@ -120,10 +125,10 @@ def initialize_detection_tables(detected_table, true_table,
     true_table['true_id'] = range(num_true)  # id in blend [0-num_true]
     # index of blend in test size [0 - len(blend_summary)]
     true_table['blend_index'] = np.ones(
-        num_true, dtype=int)*(batch_index*batch_size + blend_index)
+        num_true, dtype=int) * (batch_index * batch_size + blend_index)
     # index of batch in test size [0 - batch_size]
     true_table['batch_index'] = np.ones(
-        num_true, dtype=int)*batch_index
+        num_true, dtype=int) * batch_index
     # number of times object was detected [0 - num_det]
     true_table['num_detections1'] = np.zeros(num_true, dtype=int)
     true_table['num_detections2'] = np.zeros(num_true, dtype=int)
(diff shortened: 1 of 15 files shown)


ismael2395:
Hi @sowmyakth , it seems that travis is failing on `imagecodecs` do you know about this package? Or has it given you trouble before ? (seems to be a gcc error) I didn't modify the requirements.txt so I didn't think it would fail on the build...

ismael2395:
Ok I think the issue was `scikit-image` , enforced it to be a specific version for now in the `requirements.txt`

ismael2395:
Hi @sowmyakth , I think I'm done with my code review :) Would you mind taking one last quick look? Thanks!

btk/plot_utils.py
```diff
@@ -95,6 +97,8 @@ def plot_blends(blend_images, blend_list, detected_centers=None,
         band_indices (list, default=[1,2,3]): list of length 3 with indices of
```
sowmyakth, line 97: Change band_indices  default in doctoring as well

tests/test_input.py
```diff
@@ -80,20 +80,21 @@ def check_output_file(user_config_dict, simulation):
     """Check if metrics output is correctly written to file.
 
     Args:
+        simulation:
         user_config_dict: Dictionary with information to run user defined
```
sowmyakth, line 84: Minor doc string modification needed that there is no `simulation_config_dict`. And `simulation` is a string indicating name of simulated blend scene like "two_gal" or "group"

setup.py
```diff
@@ -14,7 +14,6 @@
     # zip_safe=False,
     install_requires=[
         'lmfit',
-        'fitsio'
```
sowmyakth, line 17: Isn't fitsio a requirement of descwl?

docs/source/conf.py
```diff
@@ -69,7 +69,6 @@ def __div__(self, other):
     'astropy.table',
     'astropy.io',
     'astropy.io.fits',
-    'fitsio',
```
sowmyakth, line 72: Isn't fitsio a requirement of descwl?

btk/draw_blends.py
```diff
@@ -119,6 +124,13 @@ def run_single_band(Args, blend_catalog,
     # define temporary galsim image
     # this will hold isolated galaxy images that will be summed
     blend_image_temp = galsim.Image(np.zeros((stamp_size, stamp_size)))
+
+    # REVIEW:
+    #  * I think this makes it a little more clear that mean_sky_level is the same for all obs_cond.
+    #  * On a different note, you are deep-copying, `obs_cond` but there is a lot of redundancy in the setup for it
+    #  right? You could get way by just resetting the stamp inside obs_cond ? Although not sure if worth it.
+
+    mean_sky_level = obs_cond.mean_sky_level
```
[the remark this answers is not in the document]
ismael2395: Thanks for the comment , I removed all the #REVIEWs so I can merge but I'm making note of your comments and of these #REVIEWs for future reference. 

Sounds good, I might get rid of the deepcopy at some point.

btk/draw_blends.py
```diff
@@ -190,7 +201,13 @@ def run_mini_batch(Args, blend_list, obs_cond):
     return mini_batch_outputs
 
 
-def generate(Args, blend_genrator, observing_generator,
+# REVIEW:
+#  * This might be a good place to start our transition to pytorch, a lot of the functionality
+#  here can be automatized with torch.Dataset and include things like caching if later turn out to be useful.
+#  This will also make the running in parallel a bit smoother.
+#  * Another question I had for speedup, right now we are generating images in the cpus. But what if we
```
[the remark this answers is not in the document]
ismael2395: Thanks I will consider it for a future PR :)

btk/plot_utils.py
```diff
@@ -284,7 +284,7 @@ def plot_metrics_summary(summary, num, ax=None, wspace=0.2, skip_zero=True):
             continue
         color = ("white" if label > 50
                  else "black" if label > 0
-                 else "grey")
+        else "grey")
```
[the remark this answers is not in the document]
ismael2395: thanks for the catch! In the future I will use a code formatter, so hopefully this doesn't happen again...

btk/compute_metrics.py
```diff
@@ -239,20 +247,24 @@ def get_blend_detection_summary(true_table, det_table):
     num_undetected2 = len(np.where(true_table['num_detections2'] == 0)[0])
     num_spurious2 = len(np.where(det_table['match_true_id2'] == -1)[0])
     num_shred2 = len(np.where(true_table['num_detections2'] > 1)[0])
-    assert num_detected1+num_undetected1+num_shred1 == num_true, "Number of "\
-        "detected objects + number undetected objects must be equal to "\
-        "the total number of true objects"
-    assert num_detected2+num_undetected2+num_shred2 == num_true, "Number of "\
-        "detected objects + number undetected objects must be equal to "\
-        "the total number of true objects"
+    assert num_detected1 + num_undetected1 + num_shred1 == num_true, "Number of " \
```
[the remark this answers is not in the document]
ismael2395: ah thanks, I hadn't realized until now you were using 79 characters.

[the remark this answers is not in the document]
ismael2395: corrected it in a bunch of other places too

setup.py
ismael2395: whoops meant to put it back after realizing that

docs/source/conf.py
ismael2395: same here

tests/test_input.py
ismael2395: thanks for the catch, corrected it in three functions

btk/plot_utils.py
ismael2395: sounds good, thanks!

btk/draw_blends.py
```diff
@@ -83,8 +86,10 @@ def draw_isolated(Args, galaxy, iso_obs):
     return iso_obs
 
 
-def run_single_band(Args, blend_catalog,
-                    obs_cond, band):
+# REVIEW:
+#  In line with not depending on catalog necessary for blends, this function is almost catalog independent.
```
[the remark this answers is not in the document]
ismael2395: Ah thanks for pointing this out, the idea for this is still a bit vague so I need to think more

PR #33, opened 2020-07-18
try to fix tests also upgraded to python 3.7

ismael2395:


ismael2395:
tests pass again after upgrading to pip and loading the `uncertainties` package before the `requirements.txt`. Looks like it might get deprecated soon though

PR #38, opened 2020-07-21
will fix test_measure

### LINKED ISSUE #37: Broken test in test_input.py
`test_measure` in `test_input.py` is broken, I suspect is because the code changed and deprecated the original results with the current np.seed but not completely sure. Skipping for now since all other tests pass and will try to resolve it soon

ismael2395:
resolves #37


ismael2395:
all test pass, but not in my computer - I suppose there must be a OS difference. Content that they pass in the CI but shouldn't be system dependent? Open an issue to resolve this later

PR #41, opened 2020-08-15
Add basic WCS

thuiop:


btk/draw_blends.py
```diff
@@ -60,6 +60,35 @@ def get_size(pixel_scale, catalog, i_obs_cond):
...
+        position of the reference pixel used as the center of the affin transform for the wcs
+    shape: tuple
+        shape of the image
+    
+    Returns
+    -------
+    wcs: WCS
+    '''
+    if center_pix == None:
+        center_pix = [(s+1)/2 for s in shape]
```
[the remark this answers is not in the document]
thuiop: Not specifically, the center does not require having integer coordinates. It is set that way to correspond to be consistent with the values of dx and dy we had.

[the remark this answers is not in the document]
ismael2395: cool that makes sense

btk/draw_blends.py
```diff
@@ -60,6 +60,35 @@ def get_size(pixel_scale, catalog, i_obs_cond):
     size = np.sqrt(r_sec ** 2 + psf_r_sec ** 2) / pixel_scale
     return Column(size, name="size")
 
+def make_wcs(pix, shape, center_pix=None, center_sky=None, naxis = 2):
```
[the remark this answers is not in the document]
ismael2395: @herjy mentioned that this would be a useful feature.

btk/draw_blends.py
```diff
@@ -188,7 +217,8 @@ def run_mini_batch(Args, blend_list, obs_cond):
             )
             blend_image_multi[:, :, j] = single_band_output[0]
             iso_image_multi[:, :, :, j] = single_band_output[1]
-        mini_batch_outputs.append([blend_image_multi, iso_image_multi, blend_list[i]])
+        wcs = make_wcs(pix=Args.pixel_scale,shape=(stamp_size, stamp_size))
```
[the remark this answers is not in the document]
ismael2395: resolving this for now , we can change it later as we continue

btk/survey.py
```diff
@@ -0,0 +1,17 @@
+import descwl
+
+
+class Survey(descwl.survey.Survey):
```
ismael2395, line 4: great! I think this is probably the simplest way without changing the code significantly

btk/draw_blends.py
```diff
@@ -188,7 +216,15 @@ def run_mini_batch(Args, blend_list, obs_cond):
             )
             blend_image_multi[:, :, j] = single_band_output[0]
             iso_image_multi[:, :, :, j] = single_band_output[1]
-        mini_batch_outputs.append([blend_image_multi, iso_image_multi, blend_list[i]])
+        wcs = make_wcs(
```
ismael2395, line 219: did you mean to have it inside the for loop?

btk/draw_blends.py
```diff
@@ -61,6 +62,33 @@ def get_size(pixel_scale, catalog, i_obs_cond):
     return Column(size, name="size")
 
 
+def make_wcs(pix, shape, center_pix=None, center_sky=None, naxis=2):
```
ismael2395, line 65: maybe we can move this function to either `create_observing_generator` or `survey` ?

btk/draw_blends.py
```diff
@@ -61,6 +62,33 @@ def get_size(pixel_scale, catalog, i_obs_cond):
...
+        pix (float): pixel size in arcseconds
+        shape (tuple): shape of the image
+        center_pix (tuple): position of the reference pixel used as the center of the
+                            affine transform for the wcs.
+        center_sky (float):
+        naxis (int):
+
+    Returns:
+        wcs: WCS
+    """
```
ismael2395, line 78: where in the code are we making the choice for arcsecs?

btk/draw_blends.py
thuiop: Good point. Actually right now the WCS information is strictly the same for each band so it does not matter from which one it is taken. Depending on how we implement multi-resolution this may be relevant later ; as of now changing the "j" index by 0 would be more clear maybe ?

btk/draw_blends.py
thuiop: It's a possibility, all the information needed is available there, we would just need to extract the wcs from obs_cond at this point in the code.

btk/draw_blends.py
thuiop: Actually there is no particular choice given to the WCS ; simply we give a pixel scale of 0.2 (for LSST) which we interpret as 0.2 arcsec, and which the wcs interprets as 0.2 degrees ; now that I think of it, it should have consequences on the result we get.  I suggest we switch to degrees by dividing the pixel scale by 3600, meaning that center_sky has to be provided in degrees and use of the WCS will require inputs or outputs in degrees, as per the documentation ; I believe there is no way to make it use arcseconds itself.

thuiop: After checking, the result is way more accurate. We can also convert automatically the center_sky in degrees but the WCS will use degrees anyway.

btk/draw_blends.py
ismael2395: Yeah maybe it's best to have it outside and index by 0 to make this explicit, and maybe have a comment just saying that it's currently the same across bands.

btk/draw_blends.py
ismael2395: I see think I understand, I think it's fine if we change to use degrees throughout. I remember now that the default catalog used is originally in degrees and at various points in the code we convert to arcsecs. We can also divide all pixel scales by 3600 in `config.py`. Maybe we can leave this for the next PR where we propagate the use of wcs consistently? Because it might break some parts of the code.

btk/draw_blends.py
ismael2395: Looking at this again: 
```

        wcs = make_wcs(
            pixel_scale=Args.pixel_scale,
            center_pix=obs_cond[j].center_pix,
            center_sky=obs_cond[j].center_sky,
            projection=obs_cond[j].projection,
            shape=(stamp_size, stamp_size),
        )
```
you are not using any information from the catalog/stamp/galsim Image in creating WCS. Do you think this might change ? if not, I would suggest creating the WCS in the `create_observing_generator.py`

btk/draw_blends.py
ismael2395: > After checking, the result is way more accurate. We can also convert automatically the center_sky in degrees but the WCS will use degrees anyway.

What 'result' are you referring to?

thuiop: > What 'result' are you referring to?

dx and dy

btk/draw_blends.py
thuiop: No, I do not think we will need the galsim information or any information from the catalog. I can move it to the create_observing_generator if you prefer.

btk/draw_blends.py
ismael2395: Ok, if it's much more accurate with degrees we should definitely do that then

btk/draw_blends.py
ismael2395: Yes, that would be great, thanks!

thuiop: Maybe it would make more sense to create it directly in the survey if we go that way ?

btk/survey.py
```diff
@@ -0,0 +1,21 @@
...
+    """ Extension of the descwl survey class including information for the WCS
+        
+        Args:
+            center_pix : tuple representing the center of the image in pixels 
+            center_sky : tuple representing the center of the image in sky coordinates
+                         (RA,DEC) in arcseconds.
```
herjy, line 10: Docstring for projection and wcs please.

btk/create_observing_generator.py
```diff
@@ -9,11 +43,14 @@ def default_obs_conditions(Args, band):
         Args: Class containing parameters to generate blends
```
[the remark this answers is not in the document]
thuiop: I think Ismael has already opened an issue to remove the Args in order to get a cleaner code, so it probably will not stay there that long. I will add something to the docstring for the time being.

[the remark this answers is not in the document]
herjy: Great! If we already fix the places that need this it will make the job easier later down the line. Here, it's just giving the precise name of the class for clarity. Thanks.

ismael2395:
Thanks @herjy for reviewing and @thuiop for working on this PR! I going to merge it now...

PR #47, opened 2020-08-25
Remove Args 

ismael2395:


btk/create_blend_generator.py
```diff
@@ -1,94 +1,47 @@
...
+    def __init__(self, catalog, batch_size=8, sampling_function=None, verbose=False):
+        """Generates a list of blend catalogs of length batch_size. Each blend
+           catalog has entries numbered between 1 and max_number, corresponding
+           to overlapping objects in the blend.
+        """
+        self.catalog = catalog
+        self.batch_size = batch_size
+        self.sampling_function = self.get_sampling_function(sampling_function)
+        self.max_number = self.sampling_function.max_number
+        self.verbose = verbose
```
thuiop, line 14: `self.verbose = verbose` should be set before the sampling_function as it is used in the get_sampling_function.

btk/create_observing_generator.py
```diff
@@ -35,69 +35,81 @@ def make_wcs(
...
+                           observing_generator.
+             stamp_size: In arcseconds.
+        """
+        if survey_name not in btk.survey.surveys:
+            raise KeyError("Survey not implemented.")
+        self.survey_name = survey_name
+        self.bands = btk.survey.surveys["survey_name"]["bands"]
+        self.pixel_scale = btk.survey.surveys["survey_name"]["pixel_scale"]
+        self.stamp_size = stamp_size
+        self.obs_function = obs_function
```
thuiop, line 80: `obs_function` is currently a function as before, should it become a class like the sampling function ?

btk/create_observing_generator.py
```diff
@@ -35,69 +35,81 @@ def make_wcs(
...
-            corresponding Args.survey_name are used to create the
-            observing_generator.
+        Args:
+             obs_function: Function that outputs dict of observing conditions. If
+                           not provided, then the default `descwl.survey.Survey` values
+                           for the corresponding survey_name are used to create the
+                           observing_generator.
+             stamp_size: In arcseconds.
+        """
+        if survey_name not in btk.survey.surveys:
```
thuiop, line 74: May become an issue later if we use other image generation than the one from WLD. I am not sure whether it is actually necessary for WLD, if a custom obs_function creating a custom Survey object is used.

btk/create_blend_generator.py
ismael2395: good catch!

btk/create_observing_generator.py
ismael2395: yeah good point, I see two advantages with converting `obs_conditions` to a class, (1) code is more 'parallel' : obs_conditions are to observing generator as sampling function is to blend generator. (2) A class for obs_conditions would make more sense as it becomes more complicated e.g. for multi resolution.

ismael2395: I'll go ahead and make this change

btk/draw_blends.py
```diff
@@ -56,195 +58,100 @@ def get_size(pixel_scale, catalog, i_obs_cond):
...
+        the mini-batches are run in parallel.
+
+        Args:
+            blend_generator: Object generator to create blended object
+            observing_generator: Observing generator for each entry in batch.
+            multiprocessing: Divides batch of blends to draw into mini-batches and
+                runs each on different core
+            cpus: If multiprocessing, then number of parallel processes to run.
+
+        Yields:
```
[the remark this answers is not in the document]
ismael2395: thanks that slipped

thuiop (commented):
So users are required to provide a sampling function ? Why remove the possibility to use the default sampling function if none is provided ?

ismael2395:
> So users are required to provide a sampling function ? Why remove the possibility to use the default sampling function if none is provided ?

Thanks for the feedback on this! Since we removed `Args`, I had to specify an arbitrary default value for `stamp_size` in the `DefaultSamplingFunction` which was hidden from the user since they just pass in `sampling_function = None`. 
I thought this was too confusing (since `stamp_size` is used in the observing generator too), so it seemed cleaner to just have the user specify the sampling_function. I welcome other suggestions to improve this though!

thuiop:
I see. Maybe they could provide a stamp_size argument which would be used for the default sampling function if no custom function is provided ? I don't know if that is really less confusing.

btk/create_observing_generator.py
ismael2395: Good point, I was thinking we can create different classes for other types of observing conditions that not necessarily return a `descwl.survey.Survey` object.

ismael2395:
> I see. Maybe they could provide a stamp_size argument which would be used for the default sampling function if no custom function is provided ? I don't know if that is really less confusing.

 @thuiop  -Thanks, that sounds a bit confusing to me as well since not all custom sampling functions might necessarily take in `stamp_size`. I think I'll leave it so users need to specify the sampling function explicitly for now, but we can revisit once we start using it and other things might come up.

I recently added code for a new class of observing conditions, could you take a look @thuiop and let me know if this new structure makes sense to you? Thanks! I think we are getting close to finish the refactoring for the image generation code.

thuiop:
It seems more consistent to me ! Do you want me to start rewriting the tests now or is there any important modifications remaining to be done ?

ismael2395:
@thuiop , I think you can start rewriting some of the tests! Please let me know if you have any questions as you go.

thuiop:
What about the "measurement" part ? You do not seem to have modified it yet.

btk/draw_blends.py
```diff
@@ -56,195 +58,102 @@ def get_size(pixel_scale, catalog, i_obs_cond):
...
+
+    def __next__(self):
+        """
+        Returns:
+            Dictionary with blend images, isolated object images, blend catalog,
+            and observing conditions.
+        """
         batch_blend_cat, batch_obs_cond, batch_wcs = [], [], []
-        stamp_size = np.int(Args.stamp_size / Args.pixel_scale)
+        pix_stamp_size = int(self.stamp_size / self.pixel_scale)
```
ismael2395, line 123: Hmm now that I think about it, fixing one pixel_scale for the draw_blend_generator might be a problem for multi resolution.

btk/create_observing_generator.py
```diff
@@ -1,109 +1,43 @@
...
+                             a given survey and band. If not provided, then the default
+                             `descwl.survey.Survey` values for the corresponding
+                             survey_name are used to create the observing_generator.
+             stamp_size: In arcseconds.
+        """
+        if survey_name not in btk.survey.surveys:
+            raise KeyError("Survey not implemented.")
+
+        if obs_conditions == None:
+            obs_conditions = DefaultObsConditions(survey_name,"i",stamp_size)
```
ismael2395, line 27: @thuiop , sorry I understand how this was confusing. My idea was to pass in the class (not object) of obs_conditions into the ObservingGenerator `__init__` and later in `__next__` to actually create the object and call it. But please do let me know what you think

I think this it will make it easier for multi resolution to not specify from the onset the "survey_name" and "band" like you do here.

ismael2395:
> What about the "measurement" part ? You do not seem to have modified it yet.

yup! I haven't modified it, but will do this soon

btk/create_observing_generator.py
thuiop: I think how it is right now is more consistent with how the sampling_function work for the blending generator. But I get what you mean and how I wrote it cannot be right as you need different conditions for each band ; you would need to save an obs_condition for each band. How about having a dictionnary containing the obs_conditions be created in the __init__ and then accessed in the __next__ to generate the survey ? That way, we may implement multi resolution as additionnal entries of this dictionnary ; there would be one observing_condition for each band in each survey.

btk/create_blend_generator.py
```diff
@@ -1,94 +1,43 @@
...
-                        )
-                if len(blend_catalog) > Args.max_number:
+            for i in range(self.batch_size):
+                blend_catalog = self.sampling_function(self.catalog)
+                if len(blend_catalog) > self.max_number:
                     raise ValueError(
                         "Number of objects per blend must be "
                         "less than max_number: {0} <= {1}".format(
-                            len(blend_catalog), Args.max_number
+                            len(blend_catalog), self.max_number
```
HironaoMiyatake, line 36: If I understand correctly, the limit from `max_number` should be taken care of in a `sampling_function()` defined by a user. In the exception message, it might be nice to suggest a user should fix the `sampling_function()`.

btk/survey.py
```diff
@@ -1,25 +1,41 @@
...
+            "i",
+            "r",
+            "g",
+        ),
+        "pixel_scale": 0.17,
+    },
+}
+
 
 class Survey(descwl.survey.Survey):
```
HironaoMiyatake, line 20: The name `Survey` might be confusing, since it's actually like a postage stamp because of the inclusion of WCS. Maybe `ObsPostage`, `ObsCutout`, or something is better.

btk/obs_conditions.py
```diff
@@ -0,0 +1,117 @@
...
+    w = WCS.WCS(naxis=2)
+    w.wcs.ctype = ["RA---" + projection, "DEC--" + projection]
+    w.wcs.crpix = center_pix
+    w.wcs.cdelt = [pixel_scale for _ in range(naxis)]
+    w.wcs.crval = center_sky
+    w.array_shape = shape
+    return w
+
+
+class WLDObsConditions(ABC):
```
HironaoMiyatake, line 41: Do we need `WLD` in this class name?

btk/create_blend_generator.py
```diff
@@ -1,94 +1,43 @@
...
+            sampling_function (btk.sampling_functions.SamplingFunction): An object that
+                                                                         return samples
+                                                                         from the catalog.
+            batch_size (int): Size of batches returned.
+            verbose: Whether to print additional information.
+        """
+        self.catalog = catalog
+        self.batch_size = batch_size
+        self.verbose = verbose
+        self.sampling_function = sampling_function
```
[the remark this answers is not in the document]
thuiop: We discussed a bit about it here https://github.com/LSSTDESC/BlendingToolKit/pull/47#issuecomment-683129702

btk/create_observing_generator.py
ismael2395: @thuiop thanks, I agree that it would be nice to keep consistency with how `sampling_function` works otherwise possibly causing confusion, so I struggled a bit with this change. 

Can you elaborate on your dictionary idea to make sure I'm understanding correctly? One thing is that I don't want to 'cache' the `btk.survey.Survey` from the beginning but allow it to change between calls of `__next__` (users might want different observing conditions between different batches if e.g. they are building algorithm to be robust between slightly different PSFs)

btk_input.py
```diff
@@ -117,11 +117,11 @@ def get_config_class(config_dict, catalog_name, verbose):
     return config
 
 
-def get_catalog(param, user_config_dict, selection_function_name, verbose):
+def get_catalog(user_config_dict, catalog_name, selection_function_name, verbose):
```
ismael2395, line 120: Thanks @thuiop  for starting to work on `btk_input.py` too :)

tests/test_input.py
```diff
@@ -394,15 +397,19 @@ def test_measure(input_args):
...
         user_config_dict["data_dir"], simulation_config_dict["catalog"]
     )
     # Set parameter values in param
-    param = btk_input.get_config_class(
-        simulation_config_dict, catalog_name, args.verbose
-    )
-    basic_meas(param, user_config_dict, simulation_config_dict, btk_input)
+    # param = btk_input.get_config_class(
+    #     simulation_config_dict, catalog_name, args.verbose
+    # )
```
ismael2395, line 411: I think you can remove these lines completely since `param` since these returns `btk.config.Simulation_params` which we are not using anymore.

btk/obs_conditions.py
ismael2395: I wanted to be explicit that this observing conditions are based on the `survey.Survey` object from WeakLensingDeblending, but we can change it if you think that's better. 

We might add other types of observing conditions classes later for the real galsim images for example

btk/create_blend_generator.py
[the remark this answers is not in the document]
ismael2395: That's right, let's revisit this later once we have a better sense of the overall functionality. For now it has to be specified explicitly

btk/survey.py
ismael2395: Yes I agree that the 'Survey' name might be confusing. What do you think @thuiop, @herjy ?

tests/test_input.py
thuiop: Yes, I only commented them temporarily to remember they were there ; I will remove them once I finish working on those tests.

btk/survey.py
herjy: I like `ObsCutout`

ismael2395:
@thuiop , just pushed changes to make the measurement_generator into a class

btk/draw_blends.py
```diff
@@ -257,4 +166,145 @@ def generate(Args, blend_generator, observing_generator, multiprocessing=False,
...
+            if self.verbose:
+                print("Noise added to blend image")
+            generator = galsim.random.BaseDeviate(seed=np.random.randint(99999999))
+            noise = galsim.PoissonNoise(rng=generator, sky_level=mean_sky_level)
+            blend_image_temp.addNoise(noise)
+        blend_image = blend_image_temp.array
+        return blend_image, iso_image
+
+    def run_mini_batch(self, blend_list, obs_cond):
+        """Returns isolated and blended images for bend catalogs in blend_list
```
ismael2395, line 265: This function is used for multiprocessing now, is it a problem that it's a class function so we have to share the `self` between different processes?

thuiop:
I have done most of the heavy work of updating the tests to the new objects ; here are the issues that remain : 

- The tests from tests_utils use some of the custom sampling functions which were defined in utils and are now obsolete because of the changes. I have not worked much with this part of the code, so I do not know whether they are important ; do you plan to convert them ?
- Most of the functions from btk_input used a parameter called `simulation_config_dict` containing informations extracted directly from a yaml configuration file. As I have I rewritten the functions now, I can choose either to remove it altogether, conserving only keyword arguments such as `batch_size` which the users would have to provide and thus extract from the data beforehand (adding an extra step) or remove the keyword arguments and keep only the dictionary. The former seems more in the spirit of this PR as the dictionnary is a bit like "Args" but it may change things for users (provided they use the content of btk_input...).
- Two of the tests refuse to run on my computer, namely `test_input_output` and `test_metrics` ; they both use the folder `tests/unit_test_output` which does not exist by default and disappears each time I run the tests ; is this a known issue ?
- All tests which I currently fixed still fail due to an AssertionError, meaning the code runs but the results are not the expected ones ; it might be due to a seed problem or to an error in the code, which I shall look into.

btk/create_observing_generator.py
thuiop: You mean that users would be able to change the obs_conditions while keeping the same generator ? I think it is possible right now by accessing the fields of the WLDObsConditions object (like `obs_generator.obs_conditions.survey_name` for instance) ; at each next, the obs_conditions object is called, ie the get_survey function is called, meaning a whole new survey is generated from the parameters contained in the fields of the obs_conditions object.

As for the dictionnary, I mean that for each band you need a different obs_conditions (as you have a different psf and other parameters...). Those could be stored in some kind of dictionary which would contain the obs_conditions for all the bands (and possibly all the surveys with multi resolution).

btk/draw_blends.py
thuiop: I would say no but it might be the reason why test_multiprocessing gives off an error.

btk/create_observing_generator.py
thuiop: Take a look at my last commit ; I have coded what I described above (for the default case in particular).

btk/create_observing_generator.py
```diff
@@ -23,12 +23,14 @@ def __init__(
...
-            obs_conditions = DefaultObsConditions(survey_name, "i", stamp_size)
-        self.survey_name = survey_name
+        
         self.bands = btk.survey.surveys[survey_name]["bands"]
         self.stamp_size = stamp_size
-
+        if obs_conditions == None:
+            obs_conditions = {}
+            for band in self.bands:
+                obs_conditions[band] = DefaultObsConditions(survey_name, band, stamp_size)
```
ismael2395, line 32: I think I understand your idea now @thuiop , looks good

ismael2395:
Thanks @thuiop for your work with the tests and identifying the remaining issues. 

> The tests from tests_utils use some of the custom sampling functions which were defined in utils and are now obsolete because of the changes. I have not worked much with this part of the code, so I do not know whether they are important ; do you plan to convert them ?

Yes I think so, I am working right now to convert them into the new format. 

> Most of the functions from btk_input used a parameter called simulation_config_dict containing informations extracted directly from a yaml configuration file. As I have I rewritten the functions now, I can choose either to remove it altogether, conserving only keyword arguments such as batch_size which the users would have to provide and thus extract from the data beforehand (adding an extra step) or remove the keyword arguments and keep only the dictionary. The former seems more in the spirit of this PR as the dictionnary is a bit like "Args" but it may change things for users (provided they use the content of btk_input...).

Yeah so the idea of `btk_input.py` is for users to specify everything from the `.yaml` file and then BTK runs everything from start to end using only the info in the `.yaml` file. I find the current `btk_input.py` file unnecessarily confusing and probably want to revamp at some point but I find this out of the scope of this PR. If you can find a way of minimally converting it that would be great, but if it involves a lot of work I wouldn't bother and just deactive the tests that use it. In terms of your question, it seems that keeping the 'dictionary' is the way to go then? 

> Two of the tests refuse to run on my computer, namely test_input_output and test_metrics ; they both use the folder tests/unit_test_output which does not exist by default and disappears each time I run the tests ; is this a known issue ?

This is not an issue that I was aware of actually, do these tests use `btk_input.py` ? It might be a problem related to this file, please see my answer above in relation to what to do with `btk_input.py`. 

> All tests which I currently fixed still fail due to an AssertionError, meaning the code runs but the results are not the expected ones ; it might be due to a seed problem or to an error in the code, which I shall look into.

I would appreciate that, the best solution I think (if you would like to do this) is to remove all randomness from the tests. I still get `AssertionErrors` in my computer when I try to run the master branch tests for the same reason.

btk/draw_blends.py
ismael2395: Hmm what error where you seeing?

thuiop: An assertion error ; the two arrays are mismatched, but not on all fields ; the first ones give the correct output, leading me to think there is an issue with the multiprocessing specifically.

btk/create_observing_generator.py
ismael2395: Thanks your solution should work well 👍

btk/create_blend_generator.py
ismael2395: Done, thanks!

btk/create_observing_generator.py
thuiop: As it is written right now, the user has to provide the dictionary. I think it would be easier for them to give the custom class and have the dictionary and objects created internally. But that means it would not be fully consistent with the sampling function where we give the objet directly instead of the class. Maybe we should change it for the sampling function ?

btk/draw_blends.py
ismael2395: Yeah this is quite strange, thanks! I'll think about this error and try reproducing it

ismael2395:
Hi @thuiop , I finished converting the remaining sampling functions, but I moved all of them to `sampling_functions.py`. Please let me know if anything is unclear

thuiop:
> Thanks @thuiop for your work with the tests and identifying the remaining issues. Here are some responses to the points you bring up:
> 
> > The tests from tests_utils use some of the custom sampling functions which were defined in utils and are now obsolete because of the changes. I have not worked much with this part of the code, so I do not know whether they are important ; do you plan to convert them ?
> 
> Yes I think so, I am working right now to convert them into the new format.

I have seen your commits, starting to work on it !

> > Most of the functions from btk_input used a parameter called simulation_config_dict containing informations extracted directly from a yaml configuration file. As I have I rewritten the functions now, I can choose either to remove it altogether, conserving only keyword arguments such as batch_size which the users would have to provide and thus extract from the data beforehand (adding an extra step) or remove the keyword arguments and keep only the dictionary. The former seems more in the spirit of this PR as the dictionnary is a bit like "Args" but it may change things for users (provided they use the content of btk_input...).
> 
> Yeah so the idea of `btk_input.py` is for users to specify everything from the `.yaml` file and then BTK runs everything from start to end using only the info in the `.yaml` file. I find the current `btk_input.py` file unnecessarily confusing and probably want to revamp at some point but I find this out of the scope of this PR. If you can find a way of minimally converting it that would be great, but if it involves a lot of work I wouldn't bother and just deactive the tests that use it. In terms of your question, it seems that keeping the 'dictionary' is the way to go then?

It is, I made it that way in the two last commits.
> > Two of the tests refuse to run on my computer, namely test_input_output and test_metrics ; they both use the folder tests/unit_test_output which does not exist by default and disappears each time I run the tests ; is this a known issue ?
> 
> This is not an issue that I was aware of actually, do these tests use `btk_input.py` ? It might be a problem related to this file, please see my answer above in relation to what to do with `btk_input.py`.

Yes, they are related to btk_input ; I think they are supposed to be testing whether the output of a specific test is alright, but fails to do so.
> > All tests which I currently fixed still fail due to an AssertionError, meaning the code runs but the results are not the expected ones ; it might be due to a seed problem or to an error in the code, which I shall look into.
> 
> I would appreciate that, the best solution I think (if you would like to do this) is to remove all randomness from the tests. I still get `AssertionErrors` in my computer when I try to run the master branch tests for the same reason.
I will look into that once I am finished with the sampling functions.

ismael2395:
> Edit : I made some progress, I had forgotten to modify the main from btk_input. Still, there is an issue which is not easy to fix : the code cannot run for any sampling_function in the config file, as to create the object, a variable amount of arguments must be given...
Thanks @thuiop , I figured we would encounter this problem at some point. In the future it might be nice to have this future. Does the problem go away if we make everything keyword arguments?

ismael2395:
> I am not sure how to get around all the randomness. Concerning the choice of galaxies we could reduce the catalog to use a limited amount of galaxy (or use a selection function to do that), but I do not know how we would do that for the shifts ; currently there is no avoiding randomness for the default sampling function.

There are several ways I can think to reduce randomness but not sure what would be best: 

(1) in the default sampling function you can set `max_shift=0` so we could eliminate randomness in the case of only 1 galaxy (having only 1 galaxy might be OK for some tests, then you can create a catalog of only 1 galaxy) 

(2) Save the images (and their information) produced separately and load them for some tests. This is especially useful for the detection tests I think which have 'magic numbers' I think.

ismael2395:
@thuiop and myself decided to merge this PR. All important tests are passing and this PR is already really large (and therefore hard to review). 

It is worth noting that `btk_input.py` is now broken, but will be fixed soon in a future/dedicated PR. This is separate from the actual source code so it should not impede future development. 

We will also revive some of the notebooks to showcase the new workflow. (which is just slightly different than before)

Thanks @thuiop for your help with the tests and code review! Thanks @herjy and @HironaoMiyatake for helpful suggestions!

PR #52, opened 2020-09-14
Add functionality for real galsim galaxies

ismael-mendoza:
## Intro
Hi @herjy, @HironaoMiyatake, @EiffL - I started this PR to start an implementation of "Real" Galsim galaxies (broadly defined). Since after having a meeting with Francois and Remy last Friday it became obvious to me that there will be some overlapping functionalities in the models that you all want to implement (e.g. both use COSMOS catalog). If it becomes too unwieldy I might later split this PR into two, but I think at least initially overlapping will be fruitful. 

I will give you some pointers on what you have to do to incorporate your galaxy models into BTK, please do let me know if after looking through the code if you have additional questions or suggestions. I'd be happy to help and make BTK better based on your experience! 

## Details
The bulk of what you have to implement will be the `run_mini_batch` function which is a method of `GalsimRealDraw` and   `GalsimHubDraw` inherited from the `DrawBlendsGenerator` abstract parent class. In essence, the `run_mini_batch` function is simple, it takes in two arguments: 

- `blend_list`: a list of catalogs, where each catalog in the list a subset of a larger catalog (in your case the COSMOS catalog) representing the galaxies to draw in the blend. So each catalog in the list corresponds to a separate postage stamp and blend. I believe you can slice a galsim COSMOS catalog? 

- `obs_conds`: A list of size equal to `blend_list` containing lists with `btk.cutout.Cutout` objects for each band in a given survey. The cutout objects contain any information you might need to draw the galaxies from the COSMOS catalog in a postage stamp. This might include WCS, PSF, stamp_size, pixel_scale, etc. You can take a look at the `btk.cutout.Cutout` class for an example of what that might include. 

And returns a list of `mini_batch_outputs` which includes the postage stamps with the blends and isolated images.

Once you identify what additional information you need for your `run_mini_batch` function to work, I can help write a separate `btk.cutout` adapted to your needs. 

Please let me know if you have any questions.


ismael2395:
@herjy I moved some of your relevant code from the other branch we were working (which I will close) on to this one. Please let me know if I missed anything

btk/galsim_draw.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+    pix_hr = hr_dict["pixel"]
+
+    lr = 0
+    hr = 0
+    loc = []
+    for i in range(n_gal):
+        k = np.int(np.random.rand(1) * len(cat))
+        shift = (np.random.rand(2) - 0.5) * shape_hr * pix_hr / 2
+        ihr, ilr, phr, plr, _ = mk_sim(
```
HironaoMiyatake, line 101: Does `mk_sim` in this line call the one in line 10? It seems the number of arguments is not consistent.

btk/obs_conditions.py
```diff
@@ -114,6 +114,34 @@ def get_psf_sky(self, psf_stamp_size):
...
+            use_true_center=True,
+            scale=self.pix,
+        ).array
+
+        # Make sure PSF vanishes on the edges of a patch that
+        # has the shape of the initial npsf
+        psf = psf - psf[0, int(self.psf_size / 2)] * 2
+        psf[psf < 0] = 0
+        psf = psf / np.sum(psf)
+
```
herjy, line 141: Careful here, we'll want to have the `InterpolatedImage` too for the convolution to go well.

herjy, line 141:
```suggestion
## Make sure PSF vanishes on the edges of a patch that has the shape of the initial npsf
            
            
            psf_obj = galsim.InterpolatedImage(galsim.Image(psf), scale=self.pix).withFlux(1.)

            ## Interpolate the new 0-ed psf
            psfs_obj.append(psf_obj)

            ## Re-draw it (with the correct fulx)
            psfs.append(psf_obj.drawImage(nx=npsf,
                                          ny=npsf,
                                          method = 'no_pixel',
                                          use_true_center = True,
                                          scale = self.pix).array)
```

herjy: I think the draw has to be with no pixel here, because pixel integration already happened.

btk/obs_conditions.py
```diff
@@ -114,6 +114,34 @@ def get_psf_sky(self, psf_stamp_size):
...
+    def psf_function(r):
+        # usually r = 0.3 * pix
+        return galsim.Moffat(2, r)
+
+    def get_psf(self, psf_stamp_size):
+        assert psf_stamp_size % 2 == 1
+        psf_int = self.psf_function(self.psf_args).withFlux(1.0)
+
+        # Draw PSF
+        psf = psf_int.drawImage(
```
herjy, line 128: Be careful about pixel integration. Make sure that it is done only once, either here or during the drawing of the final image.

btk/galsim_draw.py
```diff
@@ -0,0 +1,223 @@
+import galsim
```
herjy, line 1: This whole piece of code comes from my multi-resolution simulations. It needs to be updated by me this week. and it's actually redundant with what's in draw_blends.

btk/sampling_functions.py
```diff
@@ -48,57 +53,67 @@ def __init__(self, max_number=2, stamp_size=24.0, maxshift=None):
...
-            blend_catalog["dec"] > self.stamp_size / 2.0
+        if np.any(blend_table["ra"] > self.stamp_size / 2.0) or np.any(
+            blend_table["dec"] > self.stamp_size / 2.0
         ):
             warnings.warn("Object center lies outside the stamp")
-        return blend_catalog
+        return blend_table
 
 
 class BasicSamplingFunction(SamplingFunction):
```
herjy, line 106: Docstring would help

herjy (commented):
I need to work on the galsim image generation to clean that up and test it (at least a notebook). I'm not sure what the best way to go about dealing with catalogs is. Thinking of other kinds of models as with François's models, I wonder if we should rather expand on the sampling function class and have method that allow to pull information from the catalog. This would be up to the user to code a different `SamplingFunction` with adequate `@abstractmethod`s to pull information from the catalog. Of course we can code our own inherited classes to handle Cosmos and catsim, but I'm not sur that requiring that catalogs be made uniform is the right way to think of this.

aboucaud (commented):
I am probably not the best judge since this is my first time contributing but I feel this PR is huge in terms of modifications. And probably lacks some tests.  I have spotted some inconsistencies and wrote a comment each time.
I have at this point only reviewed the code but haven't tested it. 
But tomorrow @ismael-mendoza I can guide you through my review and you'll probably answer most of the comments straight away.

btk/draw_blends.py
```diff
@@ -227,6 +233,7 @@ def render_mini_batch(self, blend_list, cutouts, survey):
             dx, dy = get_center_in_pixels(blend_list[i], wcs)
             blend_list[i].add_column(dx)
             blend_list[i].add_column(dy)
+            # TODO: How to get size for COSMOS?
```
thuiop, line 236: That's an issue I encountered too when implementing COSMOS galaxy drawing ; we should check whether the needed parameters are given in the catalog or if we can compute them ourselves. If not, maybe consider to make it optional, so that users may use the MeasureGenerator without computing the size ?

ismael-mendoza:
BTK meeting notes, December 1, 8am EST:

Present:
Ismael Mendoza
Thomas Sainrat
Alexandre Boucault
Hironao Miyatake
Bastien Arcelin
Rémy Joseph
Cécile Roucelle

Discussion: 

- We went through Alexandre/Thomas/Remy 's review comments
- Debated whether having BlendGenerator as a Class is useful at all
- We decided to keep the Catalog object class but be made more explicit with a `get_raw_catalog` method as Alexandre proposed. 
- We debated whether the selection function was useful at all. 
- Decided to add a `get_raw_catalog` method in the Catalog class, get rid of `self.cat` property
- Thomas pointed out that assuming objects have a measure of size in every case is problematic in the Measurement Code 
- Probably we don't need abstractmethod for the `compatible_catalogs` property in DrawBlendsGenerator

Assignments: 

- Ismael will work on the Cosmos Cutout object information so Remy can use it to draw blends. 
- Alexandre will work on standarizing the catalog object 
- Remy will work on Cosmos draw blend generator and create a notebook demonstrating usage. 
- Thomas will look into the "size" problem and see if there is an easy resolution. 
- Hironao will help test Remy's notebook

New issues: 
- Alexandre proposed to create an issue related to returning things in batches for ML purposes. This might be related to #1  
- Alexandre proposed an issue to get rid of the if/else statement in the BlendGenerator.

thuiop:
Regarding the "size" problem : I have looked into the code ; the `MeasureGenerator` part does not require the size, leaving the functions in the `compute_metric`. Among them, I believe` evaluate_detection` is the most relevant. It calls upon the `get_detection_match` function which takes as argument a table containing the true parameters of the blend, and one containing the measured parameters, and matches the entries of each table. To do that, it actually has two modes : either it uses the flat euclidean distance in pixels (matching a decteted peak to the closest true peak), or it uses a renormalized distance where the euclidean distance is divided by the size of the true galaxies. Both of those are implemented and the result ultimately contains the results both for the euclidean distance and for the renormalized distance.

This means that we can just remove the second mode to avoid needing to compute the size, if we think it is not useful. I can also implement a check to ignore the second mode if the size is not defined, or we could set the size to 1 when it cannot be computed. Finally, we might change the size to something less complicated (I am open to ideas) which could be computed/measured in the COSMOS case.

aboucaud:
Refactoring of the catalogs should be quite done, see PR #62 

Review wanted pls.

btk/galsim_draw.py
```diff
@@ -0,0 +1,223 @@
...
+from astropy import wcs as WCS
+import numpy as np
+import scarlet
+
+center_ra = 19.3 * galsim.degrees  # The RA, Dec of the center of the image on the sky
+center_dec = -33.1 * galsim.degrees
+
```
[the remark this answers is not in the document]
ismael-mendoza: This code was for Remy's reference, we will remove it once he finishes that implementation.

aboucaud (approved):
The PR looks fine to me now, appart maybe from some changes expected by @herjy tomorrow. 

The tests have pointed us towards some bugs and all now pass again, without skipping  👍

The title of the PR could probably be updated to remove the galsim_hub mention since it will be dealt with later.

Thanks for your time guys, especially @ismael-mendoza and @thuiop. I have learned a lot on the btk internals and hope to be able to contribute now a bit more often.

btk/catalog.py
```diff
@@ -0,0 +1,102 @@
...
+        selection_function: callable
+            logical selection on the catalog table columns/rows.
+
+        """
+        if not callable(selection_function):
+            raise TypeError("selection_function must be callable")
+
+        self.table = selection_function(self.table)
+        self._selection_functions.append(selection_function)
+
```
HironaoMiyatake, line 58: Do we want a method to get a list of selection functions?

btk/create_blend_generator.py
```diff
@@ -4,26 +4,26 @@ def __init__(
...
-            catalog (astropy.Table.table): CatSim-like catalog.
+            catalog (btk.catalog.Catalog): BTK Catalog object
             sampling_function (btk.sampling_functions.SamplingFunction): An object that
                                                                          return samples
                                                                          from the catalog.
             batch_size (int): Size of batches returned.
-            verbose: Whether to print additional information.
             shifts (list): Contains arbitrary shifts to be applied instead of
                            random shifts. Must be of length batch_size. Must be used
                            with ids.
```
HironaoMiyatake, line 23: "Must be used with ids" -> "Must be used with indexes" ?

btk/draw_blends.py
```diff
@@ -376,3 +387,106 @@ def render_single(self, entry, cutout, band):
                 print("Source not visible")
             entry["not_drawn_" + band] = 1
             raise descwl.render.SourceNotVisible
+
+
+class GalsimRealDraw(DrawBlendsGenerator):
```
HironaoMiyatake, line 392: Why you do not use `render_single` and `render_blend` here?

ismael-mendoza:
Hello @herjy , @thuiop , @aboucaud , @HironaoMiyatake  - I'm merging this PR now since it's been a while. There is definitely work that needs to be done (which is documented in your very helpful issues) but I think most of the functionality is there and we can work on smaller PRs to improve on this . Thanks everyone!

PR #55, opened 2020-09-14
Multi resolution

thuiop:
Implementing the multi-resolution feature.
It can be activated by passing the argument `multiresolution=True` to the `ObservingGenerator`, as well as making `survey_name` a list of survey names instead of a single survey.
As a result, the outputs of the drawblendgenerator become dictionaries, with keys corresponding to each survey name provided, and containing the outputs as before. For instance, one would access the blended images from HSC with `output['blend_images']['HSC']`, which is exactly what would be in `output['blend_images']` in the single resolution case.
If `multiresolution=False` (default case), everything works as before ; therefore, existing code should not be affected.

Docstring and tests remain to be written, which will be done if no major changes are to be made.

### DIFF
--- a/btk/create_observing_generator.py
+++ b/btk/create_observing_generator.py
@@ -18,18 +18,13 @@ class ObservingGenerator:
                              `descwl.survey.Survey` values for the corresponding
                              survey_name are used to create the observing_generator.
         """
-        self.multiresolution = multiresolution
-        if multiresolution:
-            self.survey_name = survey_name
-            self.verbose = verbose
-
+        if type(survey_name) == str:
+            self.surveys = [survey_name]
+        elif type(survey_name) == list:
+            self.surveys = survey_name
         else:
-            if survey_name not in all_surveys:
-                raise KeyError("Survey not implemented.")
-
-            # self.bands = all_surveys[survey_name]["bands"]
-            self.survey_name = survey_name
-            self.verbose = verbose
+            raise TypeError("survey_name is not in the right format")
+        self.verbose = verbose
 
         # create default observing conditions
         if obs_conds is None:
@@ -42,16 +37,10 @@ class ObservingGenerator:
         return self
 
     def __next__(self):
-        if self.multiresolution:
-            observing_generator = {}
-            for s in self.survey_name:
-                observing_generator[s] = []
-                for band in all_surveys[s]["bands"]:
-                    cutout = self.obs_conds(s, band)
-                    observing_generator[s].append(cutout)
-        else:
-            observing_generator = []
-            for band in all_surveys[self.survey_name]["bands"]:
-                cutout = self.obs_conds(self.survey_name, band)
-                observing_generator.append(cutout)
+        observing_generator = {}
+        for s in self.surveys:
+            observing_generator[s] = []
+            for band in all_surveys[s]["bands"]:
+                cutout = self.obs_conds(s, band)
+                observing_generator[s].append(cutout)
         return observing_generator
--- a/btk/create_observing_generator.py
+++ b/btk/create_observing_generator.py
@@ -51,7 +51,7 @@ class ObservingGenerator:
                     observing_generator[s].append(cutout)
         else:
             observing_generator = []
-            for band in self.bands:
+            for band in all_surveys[self.survey_name]["bands"]:
                 cutout = self.obs_conds(self.survey_name, band)
                 observing_generator.append(cutout)
         return observing_generator
--- a/btk/create_observing_generator.py
+++ b/btk/create_observing_generator.py
@@ -18,12 +18,18 @@ class ObservingGenerator:
                              `descwl.survey.Survey` values for the corresponding
                              survey_name are used to create the observing_generator.
         """
-        if survey_name not in all_surveys:
-            raise KeyError("Survey not implemented.")
+        self.multiresolution = multiresolution
+        if multiresolution:
+            self.survey_name = survey_name
+            self.verbose = verbose
 
-        self.bands = all_surveys[survey_name]["bands"]
-        self.survey_name = survey_name
-        self.verbose = verbose
+        else:
+            if survey_name not in all_surveys:
+                raise KeyError("Survey not implemented.")
+
+            # self.bands = all_surveys[survey_name]["bands"]
+            self.survey_name = survey_name
+            self.verbose = verbose
 
         # create default observing conditions
         if obs_conds is None:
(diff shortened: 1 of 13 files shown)


btk/create_observing_generator.py
```diff
@@ -17,12 +18,18 @@ def __init__(
                              `descwl.survey.Survey` values for the corresponding
                              survey_name are used to create the observing_generator.
         """
-        if survey_name not in all_surveys:
-            raise KeyError("Survey not implemented.")
+        self.multiresolution = multiresolution
+        if multiresolution:
+            self.survey_name = survey_name
+            self.verbose = verbose
```
ismael2395, line 24: don't we want to allow multiple surveys ?

btk/draw_blends.py
```diff
@@ -124,46 +128,90 @@ def __next__(self):
...
-        )
-        isolated_images = np.zeros(
-            (
-                self.batch_size,
-                self.max_number,
-                pix_stamp_size,
-                pix_stamp_size,
-                len(self.bands),
+        if self.multiresolution:
+            batch_blend_cat, batch_obs_cond, batch_wcs = {},{},{}
```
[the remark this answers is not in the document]
ismael2395: I think making these a dictionaries with survey as key is a good idea

btk/create_observing_generator.py
```diff
@@ -34,8 +41,16 @@ def __iter__(self):
...
 
     def __next__(self):
-        observing_generator = []
-        for band in self.bands:
-            btk_survey = self.obs_conds(self.survey_name, band)
-            observing_generator.append(btk_survey)
+        if self.multiresolution:
+            observing_generator = {}
+            for s in self.survey_name:
+                observing_generator[s] = []
```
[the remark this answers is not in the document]
ismael2395: it's a little cumbersome to have to remember the order of the bands, another option is to store the `band` and `survey_name` inside the `cutout` object.

ismael2395 (changes requested):
Hi @thuiop , please let me know your thoughts on these comments. I also sent you a message in slack

btk/create_observing_generator.py
```diff
@@ -7,6 +7,7 @@ def __init__(
         survey_name,
         obs_conds=None,
         verbose=False,
+        multiresolution=False
```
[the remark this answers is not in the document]
thuiop: In my original design I had wondered whether a flag would be useful and finally kept it but in retrospect I think it can be removed and replaced by a check on the nature of survey_name

I heard about some Rubin-Euclid-Roman joint analysis so you never know.

btk/create_observing_generator.py
```diff
@@ -17,12 +18,18 @@ def __init__(
...
+        if multiresolution:
+            self.survey_name = survey_name
+            self.verbose = verbose
 
-        self.bands = all_surveys[survey_name]["bands"]
-        self.survey_name = survey_name
-        self.verbose = verbose
+        else:
+            if survey_name not in all_surveys:
+                raise KeyError("Survey not implemented.")
```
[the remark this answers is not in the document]
thuiop: This works too, I did not consider this check very much in the first go.

btk/create_observing_generator.py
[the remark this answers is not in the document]
ismael2395: haha ok, we can keep the `len` unconstrained then

btk/create_observing_generator.py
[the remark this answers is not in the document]
thuiop: It can probably be done but I am not sure whether there is a real benefit ? The bands are in the same order as in the images. It would make sense to the band stored in the cutout too, but I am not sure if the related code will be more readable.

btk/draw_blends.py
```diff
@@ -22,15 +22,13 @@ def get_center_in_pixels(blend_catalog, stamp_size, pixel_scale):
...
+        wcs: astropy.wcs.WCS object corresponding to the image
     Returns:
         `astropy.table.Column`: x and y coordinates of object centroid
     """
-    center = (stamp_size / pixel_scale - 1) / 2
-    dx = blend_catalog["ra"] / pixel_scale + center
-    dy = blend_catalog["dec"] / pixel_scale + center
+    dx, dy = wcs.all_world2pix(
+        blend_catalog["ra"] / 3600, blend_catalog["dec"] / 3600, 0
+    )
```
ismael2395, line 31: question which is not clear to me: now that we have WCS might it be easier to not convert 'ra' and 'dec' to arcseconds to begin with? (this is done in the sampling functions). Do you see other obstacles Thomas?

btk/draw_blends.py
```diff
@@ -278,38 +293,45 @@ def run_mini_batch(self, blend_list, obs_conds):
...
             `numpy.ndarray` of blend images and isolated galaxy images, along with
             list of blend catalogs.
         """
         mini_batch_outputs = []
         for i in range(len(blend_list)):
-            dx, dy = get_center_in_pixels(
-                blend_list[i], self.stamp_size, self.pixel_scale
-            )
+            pixel_scale = obs_conds[0].pixel_scale
+            dx, dy = get_center_in_pixels(blend_list[i], obs_conds[0].wcs)
```
ismael2395, line 306: Pixel_scale/WCS are always the same for a given survey right? We might want to restructure the cutout class or something to make this explicit.

btk/draw_blends.py
```diff
@@ -237,7 +252,7 @@ def run_single_band(self, blend_catalog, obs_conds, band):
         galaxy_builder = descwl.model.GalaxyBuilder(
             obs_conds, no_disk=False, no_bulge=False, no_agn=False, verbose_model=False
         )
-        pix_stamp_size = np.int(self.stamp_size / self.pixel_scale)
+        pix_stamp_size = np.int(self.stamp_size / obs_conds.pixel_scale)
```
ismael2395, line 255: to be consistent with naming, we might want to make the argument passed into `run_single_band` be `cutout` instead of `obs_conds`.

btk/draw_blends.py
```diff
@@ -278,38 +293,45 @@ def run_mini_batch(self, blend_list, obs_conds):
             blend_list: List of catalogs with entries corresponding to one blend.
             obs_conds (list): List of `descwl.survey.Survey` class describing
                 observing conditions in different bands.
```
ismael2395, line 295: Could you please update this docstring? I guess now a list of `btk.cutout.Cutout` objects

ismael2395 (changes requested):
A couple of minor comments, overall it's looking really good

btk/draw_blends.py
thuiop: I think the core of it was to have the pixel scale and stamp size in arcseconds, which are more adapted to the weak lensing case as the galaxy studied have a size of a few arcseconds at most, which is why I kept as much arcseconds as possible (also there might be some galsim features which require arcseconds ?). The ideal case would be to have astropy wcs support arcsecs but that is a bit out of the scope of btk.

btk/draw_blends.py
thuiop: I thought about that too, it may be clearer.

btk/draw_blends.py
ismael2395: Got it, thanks. I agree that probably better to still do the conversion of arcseconds at the beginning then.

btk/draw_blends.py
ismael2395: Hmm yeah, I agree that probably better off just having different surveys for these cases. Let me take a stab at the code to see if I can simplify the logic a little.

btk/create_observing_generator.py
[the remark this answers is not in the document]
ismael2395: You are right, I think this is fine as it is for now

ismael2395 (approved):

ismael2395:
Merging now, great work @thuiop ! 

Will leave exemplifying how this feature can be used to future PRs.

Issue #58, opened 2020-09-30
Allow for custom survey input


thuiop:
As of now, the available surveys are taken from the `all_surveys` variable in `obs_conditions.py` ; the users cannot give a new survey as an input, even if they use custom obs conditions and blend generator.

PR #59, opened 2020-09-30
Custom survey input

thuiop:
Fixes the issue #58 

Instead of storing the survey names, we now store the dictionary corresponding to the survey, either taken from `all_surveys` if the name of the survey is provided or directly given by the user.


ismael2395 (approved):

ismael2395:
Looks good @thuiop please merge.

PR #61, opened 2020-11-24
Im/simplify cutouts

ismael-mendoza:
Even easier to extend cutout or rendering new images functionality


thuiop:
If I understood this correctly, the user is intended to specify only the render_single method when implementing a custom rendering system ? I can see how this simplify things for users ; however, it might cause problem in some cases where you actually benefit from drawing all the blends using common structure (I know it is the case for our implementation of real galaxy drawing as we only have a r-band image which is "rescaled" to match an expected flux for the other bands). Of course we can still overload the render_mini_batch function to get it working as before ; just making a quick note here

ismael-mendoza:
Thanks @thuiop for your suggestion - That is correct, the intention is for the user to overwrite the `render_single` method. I agree that this is probably not the perfect balance of flexibility/convenience and I'm happy to discuss ideas for improvement over our sprint. The motivation for this was that I figured it was too cumbersome for the user to have to setup all the `np.zeros` arrays and that should probably be something that should be abstracted out. 

But yeah, I figured that if a user has a more complicated blend generation procedure in mind then they can overwrite `render_blend` or `render_mini_batch`.

PR #62, opened 2020-12-01
Refactor the Catalog objects

aboucaud:
This PR aims at refactoring the current `Catalog` class to improve its modularity.

Among the changes:
- `__init__` signature changed to initialize the `Catalog` from an existing catalog-like object.
- use of a `@classmethod` to construct the `Catalog` from a file
- the raw catalog is now a private class attribute and should be accessed through the `get_raw_catalog()` method
- selection functions on the table can be passed through the `apply_selection_function()` method and are stored for inspection

### DIFF
--- a/btk/catalog.py
+++ b/btk/catalog.py
@@ -6,72 +6,88 @@ import astropy.table
 
 
 class Catalog(ABC):
-    def __init__(self, catalog_file, selection_function=lambda x: x, verbose=False):
+    def __init__(self, catalog, verbose=False):
         """Returns astropy table with catalog name from input class.
 
         Args:
-            catalog_file: File path of CatSim-like catalog or galsim COSMOS catalog to
-                         draw galaxies from.
-            selection_function: Selection cuts (if input) to place on input catalog.
+            catalog : CatSim-like catalog or galsim COSMOS catalog to draw galaxies from.
             verbose: Whether to print information related to loading catalog.
 
         Attributes:
-            self.cat (varies): Catalog object of varying type. Might be necessary to
-                                ultimately generate the images (e.g. COSMOS Catalog)
-            self.table (`astropy.table`): CatSim-like catalog with a selection
-                                           criteria applied if provided.
+            self.table (`astropy.table`): CatSim-like catalog with selection criteria applied 
+                and recorded in the `_selection_functions` list.
         """
-        self.cat = self.get_catalog(catalog_file)
-        self.table = selection_function(self.get_table())
+        self._raw_catalog = catalog
         self.verbose = verbose
+        self.table = self._prepare_table()
+        self._selection_functions = []
 
         if self.verbose:
             print("Catalog loaded")
 
     @abstractmethod
-    def get_catalog(self, catalog_file):
+    @classmethod
+    def from_file(cls, catalog_file, verbose):
+        """Catalog constructor from input file"""
         pass
 
     @abstractmethod
-    def get_table(self):
+    def _prepare_table(self):
+        """Operations to standardize the catalog table"""
         pass
 
-    @abstractmethod
     @property
     def name(self):
-        # class name
-        pass
+        return self.__class__.__name__
 
+    def get_raw_catalog(self):
+        return self._raw_catalog
+
+    def apply_selection_function(self, selection_function):
+        """Apply a selection cut to the current table.
+
+        Parameters
+        ----------
+        selection_function: callable 
+            logical selection on the catalog table columns/rows.
+
+        """
+        if not callable(selection_function):
+            raise TypeError("selection_function must be callable")
+
+        self.table = selection_function(self.table)
+        self._selection_functions.append(selection_function)
 
 class WLDCatalog(Catalog):
-    def get_catalog(self, catalog_file) -> astropy.table.Table:
+    @classmethod
+    def from_file(cls, catalog_file, verbose=False):
         # catalog returned is an astropy table.
         _, ext = os.path.splitext(catalog_file)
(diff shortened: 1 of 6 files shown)


btk/catalog.py
```diff
@@ -6,66 +6,89 @@
...
         pass
 
     @property
     def name(self):
         return self.__class__.__name__
 
+    def get_raw_catalog(self):
+        return self._raw_catalog
+
+    def apply_selection_function(self, selection_function):
```
thuiop, line 46: I am still perplex about the selection_function ; shouldn't it be applied to the whole catalog in some way, and then the table gets extracted ? I do not really see an use case, aside some really basic things like slicing the catalog, which could be done in the sampling function anyway (whose job is basically to modify the table to represent the blend)

btk_input.py
```diff
@@ -157,15 +157,17 @@ def get_catalog(user_config_dict, catalog_name, selection_function_name, verbose
     else:
         utils_filename = os.path.join(os.path.dirname(btk.__file__), "utils.py")
         selection_function = None
-    catalog = btk.catalog.load_catalog(
-        catalog_name, selection_function=selection_function
-    )
+    catalog = btk.catalog.WLDCatalog.from_file(catalog_name)
```
thuiop, line 160: Thanks for updating the btk_input file, I am not sure it actually worked before though, it will need some working over (later).

thuiop (commented):
Nice PR, the Catalog class should definitely be clearer now.

btk_input.py
aboucaud: Sure, it needs to be incorporated in the module for easier use.

btk/catalog.py
aboucaud: The way I designed it, the `table` attribute is a sliced view of the raw catalog. Therefore, you can always come back to the original catalog. Whereas if you apply cuts to the catalog itself the changes are lost.

The other confusion you mention apparently comes from vocabulary when we take a _sampling function_ for a _selection function_. The way I understand it is that a sampling function is a carefully chosen set of selection functions. 
Do I understand correctly ?

thuiop: When we talk about sampling function it refers to the SamplingFunction object (which used to be a function but now is a callable object). It is where the details of the blends (for instance the positions of the galaxies etc) are defined, by modifiying the astropy table. As of now, I do not use a selection function and do everything in the sampling function, including things like mag cut which only restrict the catalog ; which is why I wonder if the selection function is really needed.

btk/catalog.py
```diff
@@ -6,66 +6,89 @@
...
-            self.cat (varies): Catalog object of varying type. Might be necessary to
-                                ultimately generate the images (e.g. COSMOS Catalog)
-            self.table (`astropy.table`): CatSim-like catalog with a selection
-                                           criteria applied if provided.
+            self.table (`astropy.table`): CatSim-like catalog with selection criteria applied 
+                and recorded in the `_selection_functions` list.
         """
-        self.cat = self.get_catalog(catalog_file)
-        self.table = selection_function(self.get_table())
+        self._raw_catalog = catalog
```
ismael-mendoza, line 20: Yes this makes sense, as well as making it a private variable

btk/catalog.py
aboucaud: Well, then that is exactly what I meant by "a set of selection functions" 😄 .
But I agree, the question is, do we need the possibility to do that at the Catalog level ?
For me it makes sense because there are some selections that can be relevant based on the knowledge of the catalog but independent of the scope filled by the SamplingFunction step.

notebooks/intro.ipynb
```diff
@@ -242,7 +247,7 @@
    "source": [
     "# let's see if our selection function is making sense by doing some histograms with the provided utility function.\n",
     "labels = ['full', 'default_selection']\n",
-    "make_histograms(catalog_full['i_ab'], catalog['i_ab'], labels=labels, loc=2)\n",
+    "make_histograms(catalog_full['i_ab'], table['i_ab'], labels=labels, loc=2)\n",
```
ismael-mendoza, line 250: ah thanks for changing the notebook!

btk/catalog.py
```diff
@@ -6,66 +6,89 @@
...
+        return table
 
 
 class CosmosCatalog(Catalog):
-    def get_catalog(self, catalog_file) -> galsim.scene.COSMOSCatalog:
+    @classmethod
+    def from_file(cls, catalog_file, verbose=False):
         # This will return a COSMOSCatalog object.
-        return galsim.COSMOSCatalog(file_name=catalog_file)
+        catalog = galsim.COSMOSCatalog(file_name=catalog_file)
```
ismael-mendoza, line 86: don't you want to return the class object here like you do above?

ismael-mendoza (approved):
I like this structure a lot better, thanks @aboucaud ! Let's just figure out whether we want to keep the selection functions

btk/catalog.py
ismael-mendoza: I agree with @thuiop that everything that is done in a "selection function" can be done in a "sampling function". But I think there are also situations where doing things at a catalog level initially might make the procedure cleaner later. 

Say that a user wants blends created from random galaxies that satisfy "i_ab < 25.3", I think in this case it is overkill for the user to have to create their own sampling function class. It would be much easier if they just specify their cuts as a selection function and use a already implemented sampling function that returns random entries from a given catalog (after cuts have been applied)

btk/catalog.py
aboucaud: If you look at the line below that's exactly what I do 😃 .

ismael-mendoza: ah just missed it, thanks!

PR #66, opened 2020-12-02
Fix tests of PR#52

aboucaud:


ismael-mendoza:
cherry-picking should have been successful

PR #75, opened 2020-12-04
Real galaxies1 galsim

herjy:
Behold the first galsim real images generated with BTK!

There is still a long way to go. Now all the functionalities for generating images are there, but it looks like the draw_blend and draw_single don't talk to each other that well. Namely, the fluxes are off, as well as the. position and I had to put a few random number to generate the image at this point. I think I know how to solve most of these issues, but I think we are safe pulling this branch now and solving the rest later or even in a completely different PR.

Screen Shot 2020-12-04 at 2 42 01 AM

### DIFF
--- a/btk/catalog.py
+++ b/btk/catalog.py
@@ -1,7 +1,7 @@
 import os
 from copy import deepcopy
+import numpy as np
 from abc import ABC, abstractmethod
-import galsim
 import astropy.table
 
 
@@ -41,7 +41,7 @@ class Catalog(ABC):
     def get_raw_catalog(self):
         return self._raw_catalog
 
-    def apply_selection_function(self, selection_function):
+    def apply_selection_function(self, selection_function, *kwargs):
         """Apply a selection cut to the current table.
 
         Parameters
@@ -53,8 +53,8 @@ class Catalog(ABC):
         if not callable(selection_function):
             raise TypeError("selection_function must be callable")
 
-        self.table = selection_function(self.table)
-        self._selection_functions.append(selection_function)
+        self.table = selection_function(self.table, *kwargs)
+        self._selection_functions.append((selection_function, *kwargs))
 
 
 class WLDCatalog(Catalog):
@@ -70,33 +70,60 @@ class WLDCatalog(Catalog):
     def _prepare_table(self):
         table = deepcopy(self._raw_catalog)
 
+
         # TODO: does the WLDCatalog require the 'ra' and 'dec' columns
         # convert ra dec from degrees to arcsec in catalog.
-        if "ra" in table.columns:
-            table["ra"] *= 3600
-        if "dec" in table.columns:
+        if "ra" in table:
+            table["Ra"] *= 3600
+        if "dec" in table:
             table["dec"] *= 3600
+
+        f = self._raw_catalog["fluxnorm_bulge"] / (
+                self._raw_catalog["fluxnorm_disk"] + self._raw_catalog["fluxnorm_bulge"]
+        )
+        r_sec = np.hypot(
+            self._raw_catalog["a_d"] * (1 - f) ** 0.5 * 4.66, self._raw_catalog["a_b"] * f ** 0.5 * 1.46
+        )
+        # BTK now requires ref_mags, but WLD still wants magnitudes
+        table["ref_mag"] = self._raw_catalog["i_ab"]
+        table["btk_size"] = r_sec
+        #ADds the extra columns to both catalogs just to be sure
+        self._raw_catalog["ref_mag"] = self._raw_catalog["i_ab"]
+        self._raw_catalog["btk_size"] = r_sec
 
         return table
 
 
 class CosmosCatalog(Catalog):
     @classmethod
-    def from_file(cls, catalog_file, verbose=False):
+    def from_file(cls, catalog_files, verbose=False):
+        """
+
+        Paramters
+        ---------
+        catalog_files: list of galsim cataolgs
+        """
         # This will return a COSMOSCatalog object.
-        catalog = galsim.COSMOSCatalog(file_name=catalog_file)
+        catalog_coord = astropy.table.Table.read(catalog_files[0], 'fits')
+        catalog_fit = astropy.table.Table.read(catalog_files[1], 'fits')
+        catalog = astropy.table.hstack([catalog_coord, catalog_fit])
 
         return cls(catalog, verbose=verbose)
 
     def _prepare_table(self):
-        table = astropy.table.Table(self._raw_catalog.real_cat)
-
+        table = deepcopy(self._raw_catalog)
         # make notation for 'ra' and 'dec' standard across code.
         table.rename_column("RA", "ra")
         table.rename_column("DEC", "dec")
+        table.rename_column("MAG", "ref_mag")
 
         # convert ra dec from degrees to arcsec in catalog.
         table["ra"] *= 3600
         table["dec"] *= 3600
 
+        size = self._raw_catalog["flux_radius"]*self._raw_catalog["PIXEL_SCALE"]
+        table["btk_size"] = size
+        #ADds the extra columns to both catalogs just to be sure
+        self._raw_catalog["ref_mag"] = self._raw_catalog["MAG"]
+        self._raw_catalog["btk_size"] = size
         return table
--- a/btk/catalog.py
+++ b/btk/catalog.py
@@ -54,7 +54,7 @@ class Catalog(ABC):
             raise TypeError("selection_function must be callable")
 
         self.table = selection_function(self.table, *kwargs)
-        self._selection_functions.append((selection_function, *kwargs))
+        self._selection_functions.append((selection_function, kwargs))
 
 
 class WLDCatalog(Catalog):
@@ -87,7 +87,7 @@ class WLDCatalog(Catalog):
         # BTK now requires ref_mags, but WLD still wants magnitudes
         table["ref_mag"] = self._raw_catalog["i_ab"]
         table["btk_size"] = r_sec
-        #ADds the extra columns to both catalogs just to be sure
+        #Adds the extra columns to both catalogs just to be sure
         self._raw_catalog["ref_mag"] = self._raw_catalog["i_ab"]
         self._raw_catalog["btk_size"] = r_sec
 
--- a/btk/catalog.py
+++ b/btk/catalog.py
@@ -116,6 +116,7 @@ class CosmosCatalog(Catalog):
         table.rename_column("RA", "ra")
         table.rename_column("DEC", "dec")
         table.rename_column("MAG", "ref_mag")
+        index = np.where(t["IDENT_1"] == self.raw_catalog["IDENT_1"] for t in table)
 
         # convert ra dec from degrees to arcsec in catalog.
         table["ra"] *= 3600
@@ -123,6 +124,7 @@ class CosmosCatalog(Catalog):
 
         size = self._raw_catalog["flux_radius"]*self._raw_catalog["PIXEL_SCALE"]
         table["btk_size"] = size
+        table["btk_index"] = index
         #ADds the extra columns to both catalogs just to be sure
         self._raw_catalog["ref_mag"] = self._raw_catalog["MAG"]
         self._raw_catalog["btk_size"] = size
--- a/btk/create_observing_generator.py
+++ b/btk/create_observing_generator.py
@@ -1,4 +1,4 @@
-from btk.obs_conditions import WLDObsConditions, all_surveys
+from btk.obs_conditions import WLDObsConditions, all_surveys, Survey
 
 
 class ObservingGenerator:
@@ -12,28 +12,22 @@ class ObservingGenerator:
         """Generates class with observing conditions in each band.
 
         Args:
-             surveys (str or list): Name(s) of the survey which should be available
-                                    in descwl. May either be a string for single survey
-                                    or a list of strings for several surveys. See
+             surveys (`Survey` or `list`): btk.obs_condition.Survey object or a list of surveys surveys. See
                                     obs_conditions.py for a list of available surveys.
              obs_conds: Class (not object) that returns observing conditions for
                              a given survey and band. If not provided, then the default
                              `descwl.survey.Survey` values for the corresponding
                              survey_name are used to create the observing_generator.
         """
-        if isinstance(surveys, str):
-            self.surveys = [all_surveys[surveys]]
+        if isinstance(surveys, Survey):
+            self.surveys = [surveys]
         elif isinstance(surveys, list):
             self.surveys = []
             for s in surveys:
-                if isinstance(s, str):
-                    if s not in all_surveys:
-                        raise KeyError("Survey not implemented.")
-                    self.surveys.append(all_surveys[s])
-                elif isinstance(s, dict):
+                if isinstance(s, Survey):
                     self.surveys.append(s)
                 else:
-                    raise TypeError("surveys is not in the right format")
+                    raise TypeError("surveys should be a `btk.obs_conditions.Survey` object")
         else:
             raise TypeError("surveys is not in the right format")
 
@@ -56,8 +50,8 @@ class ObservingGenerator:
     def __next__(self):
         observing_generator = {}
         for s in self.surveys:
-            observing_generator[s["name"]] = []
-            for band in s["bands"]:
+            observing_generator[s.name] = []
+            for band in s.bands:
                 cutout = self.obs_conds(s, band)
-                observing_generator[s["name"]].append(cutout)
+                observing_generator[s.name].append(cutout)
         return observing_generator
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -51,12 +51,8 @@ def get_size(pixel_scale, catalog, cutout):
     Returns:
         `astropy.table.Column`: size of the galaxy.
     """
-    f = catalog["fluxnorm_bulge"] / (
-        catalog["fluxnorm_disk"] + catalog["fluxnorm_bulge"]
-    )
-    hlr_d = np.sqrt(catalog["a_d"] * catalog["b_d"])
-    hlr_b = np.sqrt(catalog["a_b"] * catalog["b_b"])
-    r_sec = np.hypot(hlr_d * (1 - f) ** 0.5 * 4.66, hlr_b * f ** 0.5 * 1.46)
+
+    r_sec = catalog["btk_size"]
     psf = cutout.psf_model
     psf_r_sec = psf.calculateMomentRadius()
     size = np.sqrt(r_sec ** 2 + psf_r_sec ** 2) / pixel_scale
@@ -114,7 +110,7 @@ class DrawBlendsGenerator(ABC):
 
         self.meas_bands = {}
         for i, s in enumerate(self.surveys):
-            self.meas_bands[s["name"]] = meas_bands[i]
+            self.meas_bands[s.name] = meas_bands[i]
 
         self.add_noise = add_noise
         self.min_snr = min_snr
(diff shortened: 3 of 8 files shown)


btk/create_observing_generator.py
```diff
@@ -12,28 +12,22 @@ def __init__(
...
-                        raise KeyError("Survey not implemented.")
-                    self.surveys.append(all_surveys[s])
-                elif isinstance(s, dict):
+                if isinstance(s, Survey):
                     self.surveys.append(s)
                 else:
-                    raise TypeError("surveys is not in the right format")
+                    raise TypeError("surveys should be a `btk.obs_conditions.Survey` object")
         else:
             raise TypeError("surveys is not in the right format")
```
thuiop, line 32:
```suggestion
            raise TypeError("surveys should be a `btk.obs_conditions.Survey` object")
```

btk/draw_blends.py
```diff
@@ -322,7 +319,7 @@ def render_blend(self, blend_catalog, cutout, band):
                 print("Noise added to blend image")
             generator = galsim.random.BaseDeviate(seed=np.random.randint(99999999))
             noise = galsim.PoissonNoise(rng=generator, sky_level=mean_sky_level)
-            _blend_image.addNoise(noise)
+            #_blend_image.addNoise(noise)
```
thuiop, line 322: Why is this line commented ?

btk/obs_conditions.py
```diff
@@ -3,18 +3,76 @@
 import numpy as np
 import galsim
 import descwl
+from collections import namedtuple
 
+# A simple class archetype to serve as a dictionary without having to write the field names every time
+Survey = namedtuple("Survey", ["name", "pixel_scale", "psf_scale", "bands", "mean_sky_level", "exp_time", "zero_point"])
```
thuiop, line 9: Those are a nice addition. I was thinking that I have not seen any checks on whether a given survey is appropriate for a catalog (ie you could not get HST images with Catsim and WLD for instance) ; should there be some implemented somewhere in the code ? (in the DrawBlendsGenerator maybe ?)

btk/obs_conditions.py
```diff
@@ -99,51 +164,69 @@ def get_psf_sky(self, psf_stamp_size):
...
-        self.psf_scale = psf_scale
+        """
+        super(CosmosCutout, self).__init__(stamp_size, survey.pixel_scale)
+        self.survey = survey
         self.psf_stamp_size = psf_stamp_size
+        self.band = band
 
     @staticmethod
     def psf_function(r):
         return galsim.Moffat(2, r)
```
thuiop, line 193: Do you think there should be support for some other PSF profile, perhaps provided by the user ? I was thinking about having the psf_function as an attribute of the cutout, allowing users to input a custom one via modified ObsConditions.

btk/draw_blends.py
```diff
@@ -437,54 +441,27 @@ def draw_single(self, cat, shift):
         Returns:
             gal (galsim.InterpolatedImage): The galsim profile of a single galaxy
         """
-
-        k = np.int(np.random.randn(1) * cat.size)
-        gal = cat.makeGalaxy(
-            k, gal_type="real", noise_pad_size=self.stamp_size * self.pixel_scale
-        )
+        k = int(np.random.rand(1)*len(self.cat))#catalog_line["btk_index"][0]
```
thuiop, line 444: I believe the galaxy id should be decided in the sampling function (because that is where we already have taken some entries in the catalog, it does not seem right to take an entirely different profile, plus the parameters like magnitude are bound to be wrong). So I guess the commented code should be the right one.

thuiop (commented):
Overall this is a really nice PR, thank you Rémy for all the work on this issue. I have made a few comments on some details which should be looked at but otherwise I think this can be merged.

btk/obs_conditions.py
HironaoMiyatake: This would be a good addition. The most flexible PSF format would be an image itself. It would be also nice to add an functionality to talk to DM stack which enable BTK to simulate Rubin images with the PSF determined from data (and also HSC images, which is actually what I want to use :))

btk/catalog.py
```diff
@@ -70,33 +70,62 @@ def from_file(cls, catalog_file, verbose=False):
...
         table = deepcopy(self._raw_catalog)
 
+
         # TODO: does the WLDCatalog require the 'ra' and 'dec' columns
         # convert ra dec from degrees to arcsec in catalog.
-        if "ra" in table.columns:
-            table["ra"] *= 3600
-        if "dec" in table.columns:
+        if "ra" in table:
+            table["Ra"] *= 3600
```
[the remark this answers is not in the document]
ismael-mendoza: yeah I think it should be 'ra'

btk/utils.py
```diff
@@ -306,7 +306,7 @@ def make_true_seg_map(image, threshold):
     return seg_map.astype(np.bool)
 
 
-def basic_selection_function(catalog):
+def basic_selection_function(catalog, max_size = 4, max_mag = 27):
```
ismael-mendoza, line 309: way simpler, thanks. It will be easier to decide later if we want to keep selection functions

btk/obs_conditions.py
ismael-mendoza: Thanks for the suggestion, @HironaoMiyatake please go ahead and open and issue and we can all think about this next

btk/obs_conditions.py
ismael-mendoza: That's right, we should make a check where not all surveys are compatible with all DrawBlendsGenerators. Maybe add a class attribute `compatible_surveys`?

btk/draw_blends.py
```diff
@@ -288,8 +282,12 @@ def render_blend(self, blend_catalog, cutout, band):
...
 
         """
-        if not hasattr(cutout, "mean_sky_level"):
-            raise AttributeError("cutout needs a mean_sky_level as an attribute.")
+        if not hasattr(cutout, "survey"):
+            mean_sky_level = cutout.mean_sky_level
+        elif not hasattr(cutout, "mean_sky_level"):
+            mean_sky_level = cutout.survey.mean_sky_level[[b == band for b in cutout.survey.bands]]
+        else:
+            raise AttributeError("cutout needs a `survey`  as an attribute.")
```
ismael-mendoza, line 290: I think this is fine for now, ultimately we would like all cutouts to be standarized (I think this is reasonable) to the point that all of them will have the `mean_sky_level` attribute (or something along these lines to calculate a background)

btk/catalog.py
```diff
@@ -70,33 +70,62 @@ def from_file(cls, catalog_file, verbose=False):
...
+        f = self._raw_catalog["fluxnorm_bulge"] / (
+                self._raw_catalog["fluxnorm_disk"] + self._raw_catalog["fluxnorm_bulge"]
+        )
+        r_sec = np.hypot(
+            self._raw_catalog["a_d"] * (1 - f) ** 0.5 * 4.66, self._raw_catalog["a_b"] * f ** 0.5 * 1.46
+        )
+        # BTK now requires ref_mags, but WLD still wants magnitudes
+        table["ref_mag"] = self._raw_catalog["i_ab"]
+        table["btk_size"] = r_sec
+        #ADds the extra columns to both catalogs just to be sure
```
ismael-mendoza, line 90: typos

btk/obs_conditions.py
```diff
@@ -3,18 +3,76 @@
...
+             sigma_HSC, ["g", "r", "i", "z", "y"],
+             np.array([21.4, 20.6, 19.7, 18.3, 17.9]),
+             np.array([600, 600, 1200, 1200, 1200]),
+             np.array([91.11, 87.74, 69.80, 29.56, 21.53]))
+Roman = Survey("Roman",
+               pix_ROMAN,
+               sigma_ROMAN,
+               ["F062", "Z087", "Y106", "J129", "H158", "F184"],
+               np.array([22, 22, 22, 22, 22, 22]), ## Not Checked!!!
+               np.array([3000, 3000, 3000, 3000, 3000, 3000]),  ## Not Checked!!!
```
ismael-mendoza, line 52: maybe make this an issue so we don't forget that these are not checked?

ismael-mendoza: you could also raise a warning whenever this survey is used somewhere in the DrawBlendsGenerator like 
```
if survey.name == 'Roman': 
    warnings.warn("Values for mean_sky_level and exp_time for the Roman survey might be incorrect")
```

btk/catalog.py
```diff
@@ -41,7 +41,7 @@ def name(self):
     def get_raw_catalog(self):
         return self._raw_catalog
 
-    def apply_selection_function(self, selection_function):
+    def apply_selection_function(self, selection_function, *kwargs):
```
aboucaud, line 44: Is this for additional arguments or keyword arguments ?
Arguments unpacking is done with `*`, keyword args with `**` since it's a dictionary.

btk/catalog.py
```diff
@@ -53,8 +53,8 @@ def apply_selection_function(self, selection_function):
         if not callable(selection_function):
             raise TypeError("selection_function must be callable")
 
-        self.table = selection_function(self.table)
-        self._selection_functions.append(selection_function)
+        self.table = selection_function(self.table, *kwargs)
+        self._selection_functions.append((selection_function, *kwargs))
```
aboucaud, line 57: The question remains here, if you pass keyword args then you need to save the dictionary in the tuple, and not unpack it =>
`self._selection_functions.append((selection_function, kwargs))`

btk/catalog.py
aboucaud, line 77:
```suggestion
        if "ra" in table:
            table["ra"] *= 3600
```

btk/catalog.py
```diff
@@ -70,33 +70,62 @@ def from_file(cls, catalog_file, verbose=False):
     def _prepare_table(self):
         table = deepcopy(self._raw_catalog)
 
+
         # TODO: does the WLDCatalog require the 'ra' and 'dec' columns
         # convert ra dec from degrees to arcsec in catalog.
-        if "ra" in table.columns:
-            table["ra"] *= 3600
-        if "dec" in table.columns:
+        if "ra" in table:
```
aboucaud, line 76: this will raise a `FutureWarning`, you should keep `table.columns` or `table.colnames` if you wish.

btk/catalog.py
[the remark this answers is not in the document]
aboucaud:
```suggestion
            table["ra"] *= 3600
```

btk/catalog.py
```diff
@@ -70,33 +70,62 @@ def from_file(cls, catalog_file, verbose=False):
...
+            table["Ra"] *= 3600
+        if "dec" in table:
```
aboucaud, line 78:
```suggestion
        if "dec" in table.colnames:
```

btk/catalog.py
aboucaud:
```suggestion
        if "ra" in table.colnames:
```

btk/catalog.py
```diff
@@ -70,33 +70,62 @@ def from_file(cls, catalog_file, verbose=False):
...
         # convert ra dec from degrees to arcsec in catalog.
         table["ra"] *= 3600
         table["dec"] *= 3600
 
+        size = self._raw_catalog["flux_radius"]*self._raw_catalog["PIXEL_SCALE"]
+        table["btk_size"] = size
+        table["btk_index"] = index
+        #ADds the extra columns to both catalogs just to be sure
+        self._raw_catalog["ref_mag"] = self._raw_catalog["MAG"]
+        self._raw_catalog["btk_size"] = size
```
aboucaud, line 130: Since you are using many columns in the table, there's a high chance of failure if one of them is missing from the input cat.
I guess we should have a list of mandatory columns, check if they're all present and fail with a nice exception message if not. Especially if you're using case sensitive column names.

btk/catalog.py
```diff
@@ -70,33 +70,62 @@ def from_file(cls, catalog_file, verbose=False):
...
 
     def _prepare_table(self):
-        table = astropy.table.Table(self._raw_catalog.real_cat)
-
+        table = deepcopy(self._raw_catalog)
         # make notation for 'ra' and 'dec' standard across code.
         table.rename_column("RA", "ra")
         table.rename_column("DEC", "dec")
+        table.rename_column("MAG", "ref_mag")
+        index = np.where(t["IDENT_1"] == self.raw_catalog["IDENT_1"] for t in table)
```
aboucaud, line 119:
```suggestion
        index = np.where(t["IDENT_1"] == self._raw_catalog["IDENT_1"] for t in table)
```

btk/obs_conditions.py
thuiop: I have been looking into it but it seems like there are different possible exposing times ; it does seem to me that those values are placeholders so we should definitely look into it at some point.

btk/catalog.py
```diff
@@ -70,33 +70,62 @@ def from_file(cls, catalog_file, verbose=False):
...
+
+        Paramters
+        ---------
+        catalog_files: list of galsim cataolgs
+        """
         # This will return a COSMOSCatalog object.
-        catalog = galsim.COSMOSCatalog(file_name=catalog_file)
+        catalog_coord = astropy.table.Table.read(catalog_files[0], 'fits')
+        catalog_fit = astropy.table.Table.read(catalog_files[1], 'fits')
+        catalog = astropy.table.hstack([catalog_coord, catalog_fit])
```
aboucaud, line 109: I'm no fan of passing lists as arguments, especially if the order in the list matters. 
Very hard to remember which order you should put them.

Do we need the two different catalogs ? If so we should use two distinct arguments.

btk/obs_conditions.py
```diff
@@ -3,18 +3,76 @@
...
+pix_HSC = 0.167
+pix_DES = 0.263
+pix_CFHT = 0.185
+
+# Sigma of the psf profile in arcseconds.
+sigma_ROMAN = 0.11 * np.array([1.68, 1.69, 1.86, 2.12, 2.44, 2.71])  # https://arxiv.org/pdf/1702.01747.pdf Z-band
+sigma_RUBIN = np.array([0.327, 0.31, 0.297, 0.285, 0.276, 0.267])  # https://www.lsst.org/about/camera/features
+sigma_EUCLID = np.array([0.16])  # https://sci.esa.int/documents/33859/36320/1567253682555-Euclid_presentation_Paris_1Dec2009.pdf
+sigma_HST = np.array([0.074])  # Source https://hst-docs.stsci.edu/display/WFC3IHB/6.6+UVIS+Optical+Performance#id-6.6UVISOpticalPerformance-6.6.1 800nm
+sigma_HSC = np.array([0.306, 0.285, 0.238, 0.268, 0.272])  # https://hsc-release.mtk.nao.ac.jp/doc/ deep+udeep
```
aboucaud, line 24: Is there a reason for choosing numpy arrays here ?
Since these are constant values, we should rather use an immutable container like a tuple.
But since you're going for the `namedtuple`, which I think is a good idea, why not also define a `Filter` or `Bandpass` the same way you define a `Survey`, with its associated `psf_scale`, `mean_sky_level`, `exp_time` and `zero_point` ? That way each survey would have a collection of Filters.

herjy: There was in my other codes where I was doing the operations on the vectors using np functions, but that is a fair point here, probably not needed.

I like the Filter class. I can implement that, it would make things look nicer and it makes a lot of sense.

btk/catalog.py
herjy: Sadely for galsim yes, there are two catalogs that contain slighlty different informations. Though we were thinking of sort of dropping the galsim catalog. It is not actually necessary at this point and we could use the default catalog to draw coordinates and fluxes as it was done before. Galsim cats only matter for the draw single part.

btk/catalog.py
herjy: Well that's the problem with handling the catalog. You never know what columns you're gonna get and you therefore need a custom `Catalog` class for each catalog.

btk/obs_conditions.py
thuiop: That seems to me to be a good idea, I think it would fit better with the Cutout system.

btk/catalog.py
thuiop: That's true, but if users want to use a new catalog, they would need to do this work anyway ; at least for the supported catalog it saves them the trouble from creating the table themselves (and they will ultimately get the results in a standard way)

btk/catalog.py
herjy: That's right, I think I made the arguments optional afterwards.

btk/obs_conditions.py
herjy: It's still early for Roman, so I think it's fine for now.

btk/draw_blends.py
herjy: Yeah I was thinking about that too, but I was unsure about it because for the WLD the attribute comes from WLD. But for cosmos since the survey carries all the informatiion it seems weird to pull the attribute out. I think this is an issue for the Cutout refactoring.

btk/obs_conditions.py
thuiop: I agree. I will write an issue to make sure we tackle this later like Ismael said.

btk/catalog.py
herjy: But my opinion of catalog is shifting, and I actually don't think that we need to build a CosmosCatalog here, at least, not unless we want to draw HST images. If we want to generate Rubin images, we should rather use the WLD catalog.

btk/obs_conditions.py
herjy: I'll do that this afternoon after my meetings are over.

btk/obs_conditions.py
```diff
@@ -3,18 +3,76 @@
...
+
+
+Euclid = Survey("Euclid",
+                pix_EUCLID,
+                sigma_EUCLID,
+                ["VIS"],
+                np.array([22.9]),
+                np.array([2260]),
+                np.array([6.85]))
```
thuiop, line 33: I notice the Euclid survey is currently only supporting the visible band ; we should add support for NIR bands in the future (as they have a lower resolution, I think the best way to do it would be to have two separate surveys Euclid_NIR and Euclid_VIS).

btk/catalog.py
thuiop: You mean that for Rubin images we should preferably use WLD only, or is it about the Catalog objects ?

btk/obs_conditions.py
herjy: Eeeewww tricky because then we run into the issue of having different resolutions for different bands. I'd punt on that or create a different survey object.

ismael-mendoza: We can discuss this further after this PR is merged, but I think two different surveys will work well? Not sure why different resolutions in different bands is an issue (since we have WCS info)

aboucaud: Yep, attach the pixel scale to the band rather than the Survey and you should be fine.

aboucaud: OR add an Instrument layer in between but that makes it heavier

thuiop: What would be the result if we had different resolutions within the same survey ? Some kind of weirdly shaped numpy array ? Currently for one survey we get a stamp_size*stamp_size*number_of_bands numpy array for each blend (+isolated_galaxies...), and when we have different surveys we get several multiband arrays (one for each survey).

herjy: Yeah so two survey objects is the easy way to go about. I like the idea of having an instrument layer in the survey object, but then we need to be careful about multi-resolution and tracking astrometry (so the wcs needs to be matched well) and then it poses implementation issues for the deblenders that deal, or not with this. This is actually a functionality that we were thinking of adding so it's coming.

aboucaud: Actually do we need a `Survey`? For image simulation and blending purposes, only the Instrument is relevant. But it could incorporate the survey in its name e.g. `Euclid_VIS` and `Euclid_NIR`

thuiop: I am not sure what you mean but things like exposure time belong to a survey more than an instrument.

aboucaud: On the contrary, exposure time is more a feature of the bandpass (filter) than instrument. This is the reason why the images in the various bands will not end up with the same depth.

thuiop: I guess we can think about that further once this PR gets merged.

herjy: I'm with Thomas here, the exposure time depends on the survey design. Even though it might be different for each band. Anyway. I'll close this conversation and open an issue.

btk/catalog.py
herjy: To generate Rubin images, you actually want a Rubin-compatible catalog with the magnitudes in the Rubin bands, that's what matters.

herjy: I'll make an other PR for that it's a small rework.

btk/obs_conditions.py
herjy: I'm rolling back on the `Filter` objects. We will have to make distinction between the `g` band from HSC and the `g` band from Rubin for instance and that will lead to the creation of a lot of objects. Maybe there is a right way to do this, but I suggest to put it in a different PR then.

ismael-mendoza:
Hello everyone, thanks for your work in reviews in this PR. I wrote down any outstanding issues I noticed and the ones raised from your comments in #82. We should address them in a later PR but for now I just want to go ahead and merge this one.

Issue #80, opened 2020-12-04
Correct shifts and magnitudes for galsim galaxies


herjy:
Following PR #52  the shifts and fluxes of galaxies generated with cosmos are  not handeled. This requires writing the proper code for `draw_single` and check how this interfaces with the `draw_blend`. To do imminently.

mpaillassa:
It seems that for now galaxy images generated with cosmos are scaled to unit flux and are not shifted. 
Have you already a precise idea on how to do it ? We may want to use `galsim.GSObject.withFlux()` and `galsim.GSObject.shift()` but we need to define how to generate the flux and shift offset values (we may also want to record them ?).

thuiop:
What I was talking about in the meeting : in the image generation code we had, there are `galsim.Bandpass` objects (which are instantiated from specific files containing the throughput for each wavelength for a particular survey. Those are used like `gal.drawImage(bandpass,image=im)` where `gal` is a GSObject (obtained by convolving the galaxy with the PSF). Before convolving the galaxy, we also multiply it by some coefficient (the same for all filters) and the number of exposures ; I have the calculation for the said coefficient, but I am not sure which value corresponds to what exactly, so the whole thing is a bit mysterious to me. 
I am not familiar with Galsim internals so I do not know if this is really relevant but it may be interesting if we want to get realistic fluxes.

PR #84, opened 2020-12-18
Change the use of catalogs

herjy:
This is an arch-minor change that fixes the use of the catalog in the Cosmos generator. It is my fault for initially not understanding the role of the catalog correctly in BTK. Here I still use the WLD catalog to draw galaxy brightnesses. Their profiles are then drawn from the Cosmos catalog. What matter is that the blend-generator receives the WLD catalog to generate Rubin images. 
I haven't removed the `CosmosCatalog` class from the code as it could be useful to draw blends at HST resolution if we ever want to do that.

### DIFF
--- a/notebooks/Gaslim_draw.ipynb
+++ b/notebooks/Gaslim_draw.ipynb
@@ -134,17 +134,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "catalog_names = [os.path.join('/Users/remy/Desktop/LSST_Project/GalSim/examples/', \n",
-    "                            'data/COSMOS_25.2_training_sample', \n",
-    "                            'real_galaxy_catalog_25.2.fits'),\n",
-    "                 os.path.join('/Users/remy/Desktop/LSST_Project/GalSim/examples/', \n",
-    "                            'data/COSMOS_25.2_training_sample', \n",
-    "                            'real_galaxy_catalog_25.2_fits.fits')]\n",
-    "galsim_catalog = galsim.COSMOSCatalog(catalog_names[0])\n",
+    "# Change this to your own path to galsim 'real_galaxy_catalog_25.2.fits'\n",
+    "galsim path = '/Users/remy/Desktop/LSST_Project/GalSim/examples/data/COSMOS_25.2_training_sample'\n",
+    "                            \n",
+    "name = os.path.join(galsim path, 'real_galaxy_catalog_25.2.fits')\n",
+    "galsim_catalog = galsim.COSMOSCatalog(name)\n",
+    "catalog_name = os.path.join(os.path.dirname(os.getcwd()), 'data', 'sample_input_catalog.fits')\n",
     "max_number = 6 # Number of galaxies in a stamp\n",
     "batch_size = 2 # Number of stamps in a batch\n",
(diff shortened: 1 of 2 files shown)


notebooks/Gaslim_draw.ipynb
```diff
@@ -134,17 +134,16 @@
...
-    "catalog_names = [os.path.join('/Users/remy/Desktop/LSST_Project/GalSim/examples/', \n",
-    "                            'data/COSMOS_25.2_training_sample', \n",
-    "                            'real_galaxy_catalog_25.2.fits'),\n",
-    "                 os.path.join('/Users/remy/Desktop/LSST_Project/GalSim/examples/', \n",
-    "                            'data/COSMOS_25.2_training_sample', \n",
-    "                            'real_galaxy_catalog_25.2_fits.fits')]\n",
-    "galsim_catalog = galsim.COSMOSCatalog(catalog_names[0])\n",
+    "# Change this to your own path to galsim 'real_galaxy_catalog_25.2.fits'\n",
+    "galsim path = '/Users/remy/Desktop/LSST_Project/GalSim/examples/data/COSMOS_25.2_training_sample'\n",
+    "                            \n",
```
aboucaud, line 139: We might leave this for now but we could switch to using environment variables or the input YAML file for these kind of path in order to avoid leaving personal details in the main repo.

notebooks/Gaslim_draw.ipynb
```diff
@@ -217,7 +216,7 @@
    "outputs": [],
    "source": [
     "# After defining the function, you can use it like: \n",
-    "catalog.apply_selection_function(btk.utils.basic_selection_function, 1, 25)\n",
+    "catalog.apply_selection_function(btk.utils.basic_selection_function, 4, 27)\n",
```
aboucaud, line 219: Maybe write those two last arguments with their signature (i.e. as keyword args) to make the selection more explicit.
Or name these arguments beforehand so we understand what 4 and 27 stand for.

notebooks/intro.ipynb
```diff
@@ -1192,7 +1192,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.7.6"
+   "version": "3.7.3"
```
aboucaud, line 1195: Does anyone know how to avoid this recurrent tiny change in the notebooks ?
This is kind of annoying :)

aboucaud (approved):
Appart from those short comments, looks good.

notebooks/intro.ipynb
ismael-mendoza: thanks for pointing this out @aboucaud, I think at some point we can standarize the development environment for BTK (so we are all using the same version of python) and will make this issue go away. Other than that I'm not sure since I don't think nbstripout takes care of this right?

aboucaud: To my knowledge nbstripout takes care of the output, not the metadata. But there is maybe an option.
Concerning the standardisation of the dev env, it usually does not go beyond specifying the minor version e.g. Python 3.7 so we would still have the issue from time to time.

ismael-mendoza:
This is a bit old now, and the intro notebook would have changed (as well as how we read the catalog). So I'm closing for now and we can revive it (if relevant) when we fix the COSMOS catalog in #91 or #80

Issue #82, opened 2020-12-12
Various issues related to new Survey/Catalog objects


ismael-mendoza:
- [ ] Check whether a given survey is appropriate for a catalog (potentially can be added in DrawBlendsGenerator as a class attribute)

- [ ] check certain mandatory column names are present in the given catalog and fail with a nice exception if they are not

- [x] make all the `sigma_X` immutable containers rather than numpy_arrays (in `obs_conditions.py`). Although then you can't multiple them so you would need to convert them to np arrays later. 

- [x] Maybe create some `Filter` object for each Survey or something else that accounts for this functionality. @aboucaud  and @herjy had some discussion: 

> why not also define a Filter or Bandpass the same way you define a Survey, with its associated psf_scale, mean_sky_level, exp_time and zero_point ?

> I'm rolling back on the Filter objects. We will have to make distinction between the g band from HSC and the g band from Rubin for instance and that will lead to the creation of a lot of objects. Maybe there is a right way to do this, but I suggest to put it in a different PR then.

- [ ] Question: Do we need `CosmosCatalog` at all or all these Catalog objects? Related comments

> Sadely for galsim yes, there are two catalogs that contain slighlty different informations. Though we were thinking of sort of dropping the galsim catalog. It is not actually necessary at this point and we could use the default catalog to draw coordinates and fluxes as it was done before. Galsim cats only matter for the draw single part. But my opinion of catalog is shifting, and I actually don't think that we need to build a CosmosCatalog here, at least, not unless we want to draw HST images. If we want to generate Rubin images, we should rather use the WLD catalog. To generate Rubin images, you actually want a Rubin-compatible catalog with the magnitudes in the Rubin bands, that's what matters.

herjy:
One comment that, I think, addresses a few questions is that I'm not sure we need the Cosmos catalog as a BTK catalog object. We only need it to generate images of individual galaxies, but we could still use other catalogs to generate a blend parameters. I'll try that in my next PR.

Issue #74, opened 2020-12-03
Simplify the user interface


herjy:
At the moment the user interface seems somewhat heavy. A user has to define a `BlendGenerator`, an `ObservingGenerator`, a `DrawBlendGenerator`, which all have pretty similar names and pipe into each other. And this is on top of the `catalog`, `sampling_functions`, `Cutout` and `obs_conditions`. This makes a LOT of objects which role I am not completely clear about. At the very least, we should come up with better, much more explicit names for these. While I understand that these classes give us some flexibility and allow the user for quick coding of custom classes I am convinced that this can be simplified. 

In an upcoming PR, I propose to rethink the name of these classes at the very least and to think of ways of simplifying the API.

thuiop:
My take on this is that : 
- The BlendGenerator and the ObservingGenerator should be hidden in the DrawBlendsGenerator (which definitely should be renamed) ; I guess they could be instantiated in the __init__. Users would provide the sampling function and the obs conditions to this unique generator and get away with it, instead of providing them to the BlendGenerator and the ObservingGenerator and then give those to the DrawBlendsGenerator.
- Similar modifications could be made to the MeasureGenerator ; this may require some thinking over, as currently the MeasureGenerator should work either with WLD or with COSMOS, without needing to modify it.
- The duo ObsConditions-Cutout is definitely the thing I find the most confusing at the moment. I believe there should be a way to combine them into a single object, probably by "absorbing" the cutout into the obs_conditions. I can see two reasons why they are separated : first, there is one cutout per band (for each survey), and we would not want users to manually setup 6+ cutouts ; they can work with ObsConditions instead to build all the cutouts from the data they already have. Secondly, the cutouts are originally built on descwl Survey object, and in the case of WLD they inherit directly from its structure, which allows them to be fed directly to WLD when rendering.

I believe that if we can adress those issues, we would get a much clearer workflow : the user defines a sampling function (exactly as of now) and observing conditions (using the new object), gives them to the generator and get the results. I feel the measure part is in an okay place right now. We would however need to clarify the metrics part but this will come in later I guess.

thuiop:
[...]

PR #85, opened 2021-01-18
Simplify user interface

thuiop:
I attempted to simplify the user interface (issue #74), by instantiating the blend and observing generator within the draw_blends generator, and by creating the Filter object within Surveys (issue #82). My aim would be for the user to only need to provide the Survey object and disregard ObsConditions and Cutouts altogether (at least for regular use of BTK). I would have liked to remove the ObsConditions completely to keep only the Cutouts but it is impractical to do right now for differentiating the WLD and COSMOS cases, so it needs some working over.


ismael-mendoza (approved):

ismael-mendoza:
Excellent, thanks for these changes. Will merge once tests are passing

PR #86, opened 2021-01-21
First simple implementation to use a fits PSF file from disk

mpaillassa:
This is the first step toward the incorporation of real PSF with Cosmos simulations. 
For now it just enables to read a PSF file from disk at the FITS format. It must be specified to `CosmosObsConditions`.

### DIFF
--- a/btk/obs_conditions.py
+++ b/btk/obs_conditions.py
@@ -140,7 +140,7 @@ class Cutout(ABC):
         self.pix_stamp_size = int(self.stamp_size / pixel_scale)
         shape = (self.pix_stamp_size, self.pix_stamp_size)
         self.wcs = make_wcs(pixel_scale, shape, **wcs_kwargs)
-
+    
 
 class WLDCutout(descwl.survey.Survey, Cutout):
     """Extension of the descwl survey class including information for the WCS.
@@ -190,7 +190,7 @@ class CosmosCutout(Cutout):
         stamp_size,
         survey,
         band,
-        psf_file,
+        psf_file=None,
         psf_stamp_size=41
     ):
         """Class containing the necessary information to draw a postage stamp (PSF,
@@ -309,7 +309,7 @@ class WLDObsConditions(ObsConditions):
 
 class CosmosObsConditions(ObsConditions):
     
-    def __init__(self, stamp_size=24, psf_stamp_size = 8.2, psf_file=None):
+    def __init__(self, stamp_size=24, psf_stamp_size=41, psf_file=None):
         super().__init__(stamp_size, psf_stamp_size)
         self.psf_file = psf_file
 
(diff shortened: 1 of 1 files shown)


aboucaud (commented):
I am wondering if the best place for the FITS extraction is in `CosmosCutout` or should rather be in `CosmosObsConditions`.
Because the way it is right now, `psf_file` and `psf_stamp_size` are uncorrelated  in `CosmosObsConditions`. But if a `psf_file` is provided, then the `psf_stamp_size` should be computed on the stamp itself, and not passed around independently.
The way I see it, `CosmoCutout` needs either, a `psf_stamp_size` OR a `psf_stamp`, but not both. If one is passed, then the other should be `None`.

Am I right here ?

mpaillassa:
Thank you for your review! I just corrected the two errors you mentioned, but `psf_file` and `psf_stamp_size` are still uncorrelated.

To define `psf_stamp_size` by the size of the provided FITS file, the FITS extraction should then be moved in the `CosmosObsConditions` initializer ? But maybe only the size should be read at this point, because it seems weird to get the PSF here while this is still done in `get_psf()` in the case of simulated PSFs. 

About the fact that `CosmosCutout` would need either `psf_stamp_size` or `psf_stamp`, I am not sure to understand what you really mean. Do you mean that you would expect a `CosmosCutout` object to represent either a stamp (other than a PSF) or a PSF stamp (but not both simultaneously) ?

aboucaud:
I'll try to clarify my thoughts. 
`get_psf()` currently ensures that the PSF stamp has a odd number of pixels : `assert self.psf_stamp_size % 2 == 1` but in the case of the FITS file, it actually bypasses that step, since you could provide a value for `psf_stamp_size` independently from the actual PSF file. 
So either you are in the simulated case and the PSF stamp size is needed as input, or you provide the PSF stamp yourself but then you need to check that its size is actually odd.

My earlier point was that if you read the FITS file at the `ObsConditions` level, then you can just pass around the PSF array you have extracted. All the needed information is in that array, unless you want to access the FITS header to get some more info but this should not matter for the `Cutout`.

mpaillassa:
Yes I think I see.
But, why do we need to satisfy this odd size condition ? Is this so that the subtracted value in `psf = psf - psf[1, int(self.psf_stamp_size / 2)] * 2` is perfectly centered ? If so, I think we should wonder if we want to do this operation with real PSF arrays.

aboucaud:
I agree with you. 
But since I did not write that particular piece of code, I'll let others bring their input here.

Ping @ismael-mendoza @herjy @HironaoMiyatake

ismael-mendoza:
Will merge this change manually once I'm done with #87

HironaoMiyatake:
@aboucaud I am not familiar with that part of GalSim, but maybe @herjy is.

ismael-mendoza:
closing this since replaced by #92

Issue #69, opened 2020-12-02
Ensure reproducibility of image production runs


aboucaud:
Reproduce results obtained by the various BTK steps would be a real added value.

So far the introductory notebook makes use of `numpy.random.seed()` but because many generators are involved, we should probably use something more scalable like passing around a random generator like `numpy.random.RandomState()`.

thuiop:
I agree ; actually, when we overhauled BTK in August, the tests broke because they were only using a seed to get reproducible results. My take on this was to allow the users to specify some galaxy ids and shifts to get some more reproducible results, but we still rely on the seed for some things (noise in particular).

ismael-mendoza:
Reopened based on additional comments in #226  from @aboucaud

thuiop:
While working on the shear PR, I've come to realize that there are still some issues : 
- the noise_seed is set per minibatch, instead of per blend, which is probably not a good idea
- the noise realisation is dependent not only on the seed, but also on the base image ; that is two different blends will get a (partially) different noise for a same random seed. I am also not sure that this is the expected behaviour, although it might be : I believe this is caused by the fact that the Poisson noise uses the values of the pixels as expectations for the noise ; it will therefore add more noise to the galaxy itself. Is that the expected behaviour ?

thuiop:
I discussed the second issue with @aguinot ; it seems that generating the noise from the background apart from the noise on the galaxy flux may be preferable (apparently ImSim does it ?). The galaxy flux noise can be simulated by drawing it with the method `phot`, although we might be fine ignoring it completely. It would be good to have other opinions before proceding.

ismael-mendoza:
Hi @thuiop thanks for bringing this up and reopening the issue. We definitely don't want any bugs in the reproducibility aspect of the codebase. 

> the noise_seed is set per minibatch, instead of per blend, which is probably not a good idea

Yeah this makes some sense. Do you have a suggestion for how to fix this? Perhaps passing around an rng to each minibatch rather than a noise seed directly?

aguinot:
[...]

PR #87, opened 2021-01-27
Remove WeakLensingDeblending dependence 

ismael-mendoza:
# Summary

- Remove WLD dependence with a small function that draws Bulge+Disk galaxies from entry
- psf is obtained with a function in `survey.py` called `get_psf` that uses the `survey` and `filter` information. 
- No more cutouts + obs_conditions
- Some deprecated unittests are being removed
- No more selection functions (user can manage catalogs, better to narrow scope of BTK if possible, closes #73 )

Suggestions welcome!

Issue #76, opened 2020-12-04
Add functionality for realistic PSF


HironaoMiyatake:
We want to add a functionality to import realistic PSF.  We should first brainstorm how to implement this. I think we can start with the following points.
- Format: The most flexible format would be an image. We may also want a capability to talk to the DM stack directly. We could add a capability to specify which `GalSim` PSF to use, since `GalSim` also has realistic PSFs such as the Kolmogorov profile and Zernike aberration model.
- We have subclasses of `Cutout` for WLD and COSMOS real images. Should we separately implement the realistic PSF capability for these subclasses, or can we share it among subclasses? I am not familiar with WLD, but it also uses galsim, so it might be possible to go with the latter.

thuiop:
WLD takes as input a bunch of parameters corresponding to the instrument and the atmosphere and generate a PSF based on those (see their Survey class for more details) ; I do not think they support providing a custom PSF.

HironaoMiyatake:
We might not need a realistic PSF for parametric galaxy models, but if we want adding such a functionality to WLD is out of our scope?

thuiop:
For now we have considered that modifying WLD for BTK purposes was out of scope ; I suppose it could be discussed (although their PSF modeling is pretty complete, even if you cannot provide a custom PSF as an image for example).

Another issue I thought about : if you are dealing with more realistic PSF, you might want to get a different one for images within the same batch ? Currently we have the same observing conditions on the whole batch.

HironaoMiyatake:
Ismael and I plan to have a telecon dedicated for this issue after the big PR is done, so we should discuss the first point then.

I should also discuss about the second point in the telecon. It would be also nice to have such a flexibility, but I do not think it's urgent. If we want to add this flexibility, maybe we want to do the same for the sky level.

thuiop:
Following the massive changes from PR #87, we should talk a bit more about how to integrate the PSF into btk. In this PR, the PSF is managed by the `get_psf` function which is called in the DrawBlendsGenerator, which returns a list of PSF models (one for each filter). I find this unpractical, as there is no way to override this function and get a custom PSF ; also, the PSF would be the same for COSMOS and Catsim (though it is not necessarily an issue). @aboucaud also pointed out that the get_psf should at least return only one PSF given a filter, instead of a list of PSF with a survey. The possibilities I see :
[...]

Issue #91, opened 2021-02-15
Fix  COSMOS catalog + add unit tests


ismael-mendoza:
In #87 the COSMOS galaxy catalog was removed from the `__init__` of the `CosmosGenerator`. I think it should be directly included in the `self.catalog`. To do imminently after closing #87

mpaillassa:
I am willing to solve this issue but I don't fully understand the current state of COSMOS catalogs and what you mean. 
The `render_single` current code in the `CosmosGenerator` seems a bit deprecated as `len(self.catalog)` fails and `catalog.makeGalaxy` does not exist anymore. Should we have a similar function to `get_catsim_galaxy` for COSMOS ?

thuiop:
I think the problem is 
1. the catalog object is not well managed. There is a galsim object called COSMOSCatalog which is the one with the makeGalaxy method, and for some reason when we rewrote a part of  BTK it got lost somewhere. My belief is that it should be stored in our Catalog object, as the field _raw_catalog, which can be accessed with the corresponding get method, and which does not have much purpose now. (this was never propery done I think)
2. The fluxes for the galaxies are wrong, which is issue #80 . Also, the function does not use the "entry" argument as it should.

ismael-mendoza:
Hi @mpaillassa, thanks for looking into this and sorry about the confusion. This issue was more of a "note to self" since I wrote it before you were going to work on this. I don't think we ever figure out exactly a good way to approach this so we are definitely open to ideas on how to structure the COSMOS object generation. 

That being said, @thuiop is entirely correct on what seems to be current the easiest approach. Regarding `1.` it seems sensible to store the actual `COSMOSCatalog` object from galsim as a `_raw_catalog` and access it later when rendering (calling the makeGalaxy method to get the corresponding galsim object). Currently it seems like it's asking for two files as a list, but it's not clear to me what are these two files or what @herjy intended. Also the current `_raw_catalog` stored is just a copy of the astropy table before the processing in `_prepare_table`. 

Another thing, I like the idea of standardizing the format of the astropy table (the `self.table` attribute) in `_prepare_table` as @herjy has suggested via `btk_size`, `btk_index`, `ref_mag`, etc. Although I don't think they have been consistently propagated and you can feel free to change this as needed. 

[...]

PR #92, opened 2021-02-22
Use a custom psf

mpaillassa:
This PR replaces the PR #86 (which is now deprecated after the numerous changes of PR #87) about the issue #76 to incorporate the possibility to use a realistic PSF from a FITS file, or in fact any custom PSF.

As suggested by @thuiop, the PSF is now directly defined in `Filter` so that it is retrieved from each `Filter` in `render_single` instead of making calls to `get_psf` (that does not exist anymore).

The new `psf` entry in `Filter` must be a galsim object. I define default galsim PSF models made of a Kolmogorov and an optical component as it was before using a very similar function to the old `get_psf`, now called `define_psf`. The principle is that it is up to the user to use these default `Survey` and `Filter` or to define its own ones using for example custom galsim PSF models built from real images (examples of this functionality should be included in the tutorial notebooks I guess).

The advantage of the new implementation is that we do not redefine the PSF at each call of `render_single`.
One default that I see is that PSF models are defined even for the surveys that are not used (maybe it would be nice to define only the surveys (and thus the PSF models) that need to be used ?). Also, there is an unpleasant redundancy between the parameters passed to `define_psf`, which are information stored in `Survey` but repeated and not retrieved from it to be passed to `define_psf` function.


thuiop:
Concerning the repeating parameters, I believe mirror_diameter and effective_area are not used elsewhere anyway, so we can remove them from the Survey and only keep their values as arguments of define_psf.
What do you mean by "surveys that are not used" ? We aim to have (if possible) all the surveys from the file available to people ; we should aim to complete the parameters in the future.

thuiop (changes requested):
Nice PR overall, I believe it completes nicely the previous one and sets a more coherent approach to how to handle the PSF in BTK. One question I have is about custom PSF from file or other : the idea is that the user should import the file into a Galsim object beforehand ? This seems great to me, if this works I am all for it.

mpaillassa:
> Concerning the repeating parameters, I believe mirror_diameter and effective_area are not used elsewhere anyway, so we can remove them from the Survey and only keep their values as arguments of define_psf.

Yes indeed.

> What do you mean by "surveys that are not used" ? We aim to have (if possible) all the surveys from the file available to people ; we should aim to complete the parameters in the future.

I just meant that if we simulate images for one specific survey, we still call `define_psf` for all other Filter of other surveys (that won't be used) but this is not that costly anyway (and I agree that BTK should have all surveys available).

btk/draw_blends.py
```diff
@@ -187,8 +184,8 @@ def __next__(self):
         for s in self.surveys:
             pix_stamp_size = int(self.stamp_size / s.pixel_scale)
 
-            # create WCS and PSF information
-            psf = [get_psf(s, filt, self.atmospheric_model) for filt in s.filters]
+            # create WCS information
+            psf = [filt.psf for filt in s.filters]
```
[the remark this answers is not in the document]
mpaillassa: Yes this is what I thought at first glance but I have noticed that in the current code PSFs are part of the output products (is this a development feature or do we really want to output PSFs as well ?) so I have let it for the moment.

btk/survey.py
```diff
@@ -43,6 +43,54 @@
...
+    if mirror_diameter > 0:
+        mirror_area = np.pi * (0.5 * mirror_diameter) ** 2
+        area_ratio = effective_area / mirror_area
+        if area_ratio <= 0 or area_ratio > 1:
+            raise RuntimeError(
+                "Incompatible effective-area and mirror-diameter values."
+            )
+        obscuration_fraction = np.sqrt(1 - area_ratio)
+        lambda_over_diameter = 3600 * np.degrees(
+            1e-10 * _central_wavelength[filt_wavelength] / mirror_diameter
```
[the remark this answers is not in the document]
mpaillassa: Indeed it is weird it should have been at least called `filt_name`, but we can indeed directly pass the value to the function.

mpaillassa:
> Nice PR overall, I believe it completes nicely the previous one and sets a more coherent approach to how to handle the PSF in BTK. One question I have is about custom PSF from file or other : the idea is that the user should import the file into a Galsim object beforehand ? This seems great to me, if this works I am all for it.

That was my first idea. This is a "a bit of work" for the user but I think it is ok given that it is for a custom additional functionality.

btk/draw_blends.py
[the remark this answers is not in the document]
ismael-mendoza: If we include some randomization in the PSF the user might want to access it later for a given batch. Although I can't exactly think of a use case for this, feel free to remove it if you find it unnecessary @mpaillassa

btk/survey.py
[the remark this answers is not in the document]
ismael-mendoza: yeah that sounds good to me too

btk/survey.py
```diff
@@ -74,13 +122,13 @@
     name="HST",
     pixel_scale=0.06,
     effective_area=1.0,  # TODO: placeholder
-    mirror_diameter=1.0,  # TODO: placeholder
+    mirror_diameter=2.4,  # TODO: placeholder
```
ismael-mendoza, line 125: were you going to remove all `mirror_diameter` and `effective_area` from all surveys to avoid redundancy? I still see it in a couple of surveys like here, but maybe you are in the process of editing

btk/draw_blends.py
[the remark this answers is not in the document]
thuiop: Oh good point I forgot about that. This would definitely be useful if we implement randomization.

ismael-mendoza (commented):
Thanks @mpaillassa for your work on this, it's starting to look a lot better.

Have you thought about how to combine this with multiple PSFs or randomized PSF (is this useful for the study you and @HironaoMiyatake are interested in?). The reason is that I'm seeing that the PSF is now created from the beginning in the filter and it's static for every batch. A potential workaround would be to pass a function that returns a psf as a galsim  object to the filter rather than the psf itself. This would allow us to call this function once for every batch and allow for randomization or a multiple (finite set) of psfs. There might be an easier way but given that we want to store the info in the filter I'm not sure. What do you think? 

(We could also leave this for the next PR)

btk/survey.py
mpaillassa: I did not start removing them yet. This one is just a change because it failed the condition 0 <= `area_ratio` < 1. 
I think `mirror_diameter=1.0` and `effective_area=1.0` were just default values with 1.0 and 1.0 ? Now I think `mirror_diameter` is right but I am not sure for `effective_area`.

mpaillassa:
Here is an update incorporating the possibility to have multiple randomized PSFs.

The principle is that now the `psf` entry of `Filter` can be either a galsim object (by default built with `define_psf` that became `define_synthetic_psf`) or the new `make_psf` function. This function checks the values of `Filter.fwhm` and `Filter.psf_path` (the other must be None when one is set) to generate a galsim PSF model either synthetically or from a FITS image. Both these entries can be lists, in which case a random element from the list is picked. 
When generating a blend, `filt.psf` is checked to be callable or not to decide if we must simply take the galsim object or call the function.

In order to build a galsim PSF model from a FITS image I took back some old code using `galsim.InterpolatedImage` which does not really make sense because the FITS image should already be at the right pixel scale. So I think we can cast the image array to a better galsim object class but I don't really know which one should be used. 
On the contrary, I noticed that `galsim.InterpolatedImage` was primarily used to set the synthetic PSF models at the right pixel scale. Is it missing in the current synthetic PSF model implementation ?

(And there are some `else` statements that are not safe in the code I added if things are not defined properly..)

btk/draw_blends.py
```diff
@@ -187,8 +184,13 @@ def __next__(self):
...
 
-            # create WCS and PSF information
-            psf = [get_psf(s, filt, self.atmospheric_model) for filt in s.filters]
+            # make PSF and WCS
+            psf = []
+            for filt in s.filters:
+                if callable(filt.psf):
+                    psf.append(filt.psf(s, filt))
+                else:
+                    psf.append(filt.psf)
```
ismael-mendoza, line 193: maybe instead of `else` check if `felt.psf` is a galsim object and otherwise raise an error (or you could check it later since if `filt.psf` is callable you want to check it returns a galsim object too)

btk/survey.py
```diff
@@ -20,7 +22,9 @@
     "Filter",
     [
         "name",
-        "atmospheric_psf_fwhm",  # in arcseconds.
+        "psf",  # galsim psf model or function to generate it
+        "fwhm", # in arcsec: must be None is psf_path is used
```
ismael-mendoza, line 26:
```suggestion
        "fwhm", # in arcsec: must be None if psf_path is used
```

btk/survey.py
```diff
@@ -43,6 +47,84 @@
...
+def make_psf(survey, filt):
+    """Generate a PSF galsim model
+    Args:
+        survey (Survey): survey to use
+        filt (Filter): filter to use
+    Returns:
+        psf_model: galsim PSF model
+    """
+
+    assert((filt.fwhm==None) ^ (filt.psf_path==None))
```
ismael-mendoza, line 107: Does the '^' stand for xor? in that case I would just make a small suggestion:
```suggestion
    assert (filt.fwhm is None) ^ (filt.psf_path is None)
```

btk/survey.py
```diff
@@ -43,6 +47,84 @@
...
+
+    # synthetic PSF model
+    if filt.psf_path==None:
```
ismael-mendoza, line 110:
```suggestion
    if filt.psf_path is None:
```

btk/survey.py
```diff
@@ -43,6 +47,84 @@
...
+        if type(filt.fwhm)==list:
```
ismael-mendoza, line 111:
```suggestion
        if isinstance(filter.fwhm, list):
```

btk/survey.py
```diff
@@ -43,6 +47,84 @@
...
+            fwhm = rd.choice(filt.fwhm)
+        else:
+            fwhm = filt.fwhm
+        psf_model = define_synthetic_psf(survey.mirror_diameter, survey.effective_area, _central_wavelength[filt.name], fwhm)
+    # FITS image PSF model
+    else:
+        if type(filt.psf_path)==list:
```
ismael-mendoza, line 118:
```suggestion
        if isinstance(filt.psf_path, list)
```

btk/survey.py
```diff
@@ -43,6 +47,84 @@
...
+            1.0
+        )
+
+    else:
+        psf_model = atmospheric_model.withFlux(1.0)
+        
+    return psf_model
+
+
+def make_psf(survey, filt):
```
ismael-mendoza, line 98: I think how you structured this function makes sense, thank you :) See a couple of commit suggestions below

btk/survey.py
ismael-mendoza: yeah I don't have the numbers to check HST, so if you or someone else knows that would be good to replace them

ismael-mendoza (changes requested):
Thanks @mpaillassa ! I think your changes make sense and we are almost ready to merge. Please see a few small comments below. For fixing the tests, I would recommend rebasing first (since I cleaned them up a bit recently). If you have any problems fixing the tests please let me know and I can help with that.

ismael-mendoza:
> In order to build a galsim PSF model from a FITS image I took back some old code using galsim.InterpolatedImage which does not really make sense because the FITS image should already be at the right pixel scale. So I think we can cast the image array to a better galsim object class but I don't really know which one should be used.
On the contrary, I noticed that galsim.InterpolatedImage was primarily used to set the synthetic PSF models at the right pixel scale. Is it missing in the current synthetic PSF model implementation ?

This is a good question which I unfortunately not sure about, so pinging others @herjy, @aboucaud, @thuiop who might better know...

thuiop:
I believe InterpolatedImage is the right object to use her ; I think they define it as the class to use if you want a GSObject with an arbitrary profile, which is exactly our case with non-parametric PSF.

I find the make_psf function a bit weird ; what if for instance I want the PSF to be a Kolmogorov profile with a random fwhm between 1.5 and 1.7 (this is an unrealistic example) ? The fact that it relies on other parameters from the Filter object is also a bit weird to me. What I envisionned was that the user would provide a "get_psf()" function (with no arguments), which returns a PSF model as the user wants (a bit like the make_psf does but more general). To get back to my example, it would go something like this : 
`def get_psf():
    fwhm = np.random.uniform(1.5,1.7)
    return galsim.Kolmogorov(fwhm=fwhm)

f = Filter(...,psf=get_psf)`
I think this would probably kind of work as it is written now, but then the fwhm parameters and psf_path are useless when the make_psf function is not used. I notice that in the code right now the function is called with survey and filter already which might be better than having no arguments. Having a default function for importing FITS files is probably useful too.

btk/survey.py
```diff
@@ -43,6 +47,84 @@
...
+        )
+        optical_psf_model = galsim.Airy(
+            lam_over_diam=lambda_over_diameter, obscuration=obscuration_fraction
+        )
+        psf_model = galsim.Convolve(atmospheric_psf_model, optical_psf_model).withFlux(
+            1.0
+        )
+
+    else:
+        psf_model = atmospheric_model.withFlux(1.0)
```
aboucaud, line 93:
```suggestion
        psf_model = atmospheric_psf_model.withFlux(1.0)
```
I don't think you can get a string with flux 1.0 :)

btk/survey.py
```diff
@@ -43,6 +47,84 @@
...
+        
+    return psf_model
```
aboucaud, line 95: I would actually return here `psf_model.withFlux(1.0)` and remove it from the if/else statement to be more explicit that it doesn't depend on anything

btk/survey.py
aboucaud: also the XOR logic here behing the assert may need to be explicited in a comment (since it's not the most straightforward thing)

btk/survey.py
aboucaud: please specify the arguments of the method to avoid confusion
```python
define_synthetic_psf(mirror_diameter=2.4, 
                                    effective_area=1.0, 
                                    filt_wavelength=_central_wavelength["f814w"], 
                                    fwhm=0.074)
```

btk/draw_blends.py
aboucaud:
```python
for filt in s.filters:
    ...
    psf.append(filt.psf(s, filt))
```
that piece of code looks really weird to me, in the sense that reading the code does not help understanding what's going on. It's a hint that some of it should be refactored.

btk/survey.py
mpaillassa: Yes it is a XOR logic but indeed I will add a comment

btk/survey.py
mpaillassa: Indeed :)

btk/draw_blends.py
mpaillassa: Yes.. In the callable case, maybe writing it in two lines like:
`psf_object = filt.psf(s, filt)`
`psf.append(psf_object)`
makes it a bit better. But it is also weird that the function that generates the PSF model needs the `Filter` as input parameter while it is itself a `Filter` parameter..

mpaillassa:
> I believe InterpolatedImage is the right object to use here ; I think they define it as the class to use if you want a GSObject with an arbitrary profile, which is exactly our case with non-parametric PSF.
> 
> I find the make_psf function a bit weird ; what if for instance I want the PSF to be a Kolmogorov profile with a random fwhm between 1.5 and 1.7 (this is an unrealistic example) ? The fact that it relies on other parameters from the Filter object is also a bit weird to me. What I envisionned was that the user would provide a "get_psf()" function (with no arguments), which returns a PSF model as the user wants (a bit like the make_psf does but more general). To get back to my example, it would go something like this :
> `def get_psf():
> fwhm = np.random.uniform(1.5,1.7)
> return galsim.Kolmogorov(fwhm=fwhm)
> 
> f = Filter(...,psf=get_psf)`
> I think this would probably kind of work as it is written now, but then the fwhm parameters and psf_path are useless when the make_psf function is not used. I notice that in the code right now the function is called with survey and filter already which might be better than having no arguments. Having a default function for importing FITS files is probably useful too.

Yes I think it would be nice to have something like your example. Currently one would need to make a list of fwhm between 1.5 and 1.7. I just found it was consistent to have the same behavior whether we use `fwhm` or `psf_path` but it may not be a good idea.
It is also true that the `fwhm` and `psf_path` parameters are a bit weird in `Filter`, especially when `make_psf` is not used (the weirdest thing being that `fwhm` is still defined and not used in this case).

Maybe a better workaround would be to have `fwhm` and `psf_path` as input parameters of `make_psf`. The default behavior would be the same than the current one (using the defined value or using a value within the list). 
Then the user would be free to define its own PSF function as long as it returns an appropriate galsim object ? 
We could also cover more use cases in `make_psf` or provide other `make_psf`-like functions in order to, for example, draw fwhm values within a given range with a given distribution like you mention ?

btk/draw_blends.py
mpaillassa: I guess it will be refactored anyway with the changes discussed with @thuiop below

btk/survey.py
mpaillassa: Just realized I missed this one. Indeed it would be better with argument names.

thuiop:
I think those we can avoid the problem of having to create a psf function for each filter by having only one (with mirror diameter etc as arguments) and using "partial" functions, either using lambda functions eg `Filter(...,psf=lambda x: get_psf(mirror_diameter=2.4,...))` (need to check if this does not give the same value everytime) or by using the `partial` function from the functools package eg `partial(get_psf,mirror_diameter=2.4,...)` ; with both we should get a callable get_psf function using the same arguments every time for a given filter. That way the parameters only relevant to the PSF would only be given to the general get_psf function (in each filter).

aguinot:
I had a quick look to the code and it seems that it is not possible to construct a pure optical PSF, right? If this is the case I would make it possible since you have setup for surveys like Euclid or HST who doesn't observe through the atmosphere. This is very minor, it can be added later given that the main focus is probably LSST.

thuiop:
> I had a quick look to the code and it seems that it is not possible to construct a pure optical PSF, right? If this is the case I would make it possible since you have setup for surveys like Euclid or HST who doesn't observe through the atmosphere. This is very minor, it can be added later given that the main focus is probably LSST.

As it is you can, as you can provide a galsim model in the psf attribute if you wish. The define_synthetic_psf does not support that at the moment though, we may do a minor modification to allow for atmospheric_model to have some value (None) which leads to no atmospheric model being convolved with the optical one.

btk/draw_blends.py
thuiop: It should become filt.psf().

mpaillassa:
Here is an update using lambda functions to enable using any custom function to generate PSF models while having a call to this function without arguments in `draw_blends.py`.
 
In the current code all surveys are using a static PSF (defined directly with a galsim object) except for Rubin which shows how to define `Filter.psf` as a function with lambda functions. It makes no sense as it uses `get_psf` that will always generate the same PSFs but I have let it for the example. I guess we should put Rubin PSFs static by default as for other surveys and we may want to have a notebook showing how to define `Filter.psf` as a function with lambda functions.
I have also let a function `get_custom_psf` that can retrieve PSFs from FITS files but this may rather go to a notebook too.

As mentioned just above, I also added the possibility to define the PSF as an optical component solely in `get_psf` if `atmospheric_model` is set as `None`.

thuiop:
> Here is an update using lambda functions to enable using any custom function to generate PSF models while having a call to this function without arguments in `draw_blends.py`.
> 
> In the current code all surveys are using a static PSF (defined directly with a galsim object) except for Rubin which shows how to define `Filter.psf` as a function with lambda functions. It makes no sense as it uses `get_psf` that will always generate the same PSFs but I have let it for the example. I guess we should put Rubin PSFs static by default as for other surveys and we may want to have a notebook showing how to define `Filter.psf` as a function with lambda functions.
> I have also let a function `get_custom_psf` that can retrieve PSFs from FITS files but this may rather go to a notebook too.
> 
> As mentioned just above, I also added the possibility to define the PSF as an optical component solely in `get_psf` if `atmospheric_model` is set as `None`.

Great job ! I think Rubin PSF should be static too and we should write an example in the tutorial notebook to showcase how it works. The function `get_custom_psf` seems fine to me where it is, we should definitely showcase how it is used at some point but at least it is there for the users to use (I would rename it `get_psf_from_file` to have a more explicit name).

btk/draw_blends.py
```diff
@@ -187,8 +184,17 @@ def __next__(self):
         for s in self.surveys:
             pix_stamp_size = int(self.stamp_size / s.pixel_scale)
 
-            # create WCS and PSF information
-            psf = [get_psf(s, filt, self.atmospheric_model) for filt in s.filters]
+            # make PSF and WCS
+            psf = []
+            for filt in s.filters:
+                if callable(filt.psf):
+                    psf.append(filt.psf()) # generate the PSF with the provided function
```
ismael-mendoza, line 191: It might be good to check at some point that `filt.psf()` (the returned psf) is indeed a `galsim.GSObject` ?

btk/survey.py
```diff
@@ -20,14 +22,95 @@
...
+        psf_model = optical_psf_model
+    elif atmospheric_psf_model is None and optical_psf_model is None:
+        raise RuntimeError(
+            f"Neither the atmospheric nor the optical PSF components are defined."
+        )
+    
+    return psf_model.withFlux(1.0)
+
+
+def get_custom_psf(psf_dir):
```
ismael-mendoza, line 92:
```suggestion
def get_psf_from_file(psf_dir):
```

ismael-mendoza (approved):
Thanks @mpaillassa this is definitely clearer. I think this code is ready for merging, I also echo @thuiop in that a notebook would be nice to illustrate the new functionality and/or unit tests for coverage. But I'll leave it up to you if you want to do this in a future PR or the current one, and if you would like help from others in doing this (if so just post an issue to remind us). Thank you :)

btk/survey.py
```diff
@@ -20,14 +22,95 @@
...
+    """Generates a custom PSF galsim model from FITS file(s)
+    Args:
+        psf_dir (string): directory where the PSF FITS files are
+    Returns:
+        psf_model: galsim PSF model
+    """
+
+    psf_files = os.listdir(psf_dir)
+    if len(psf_files)>1:
+        psf_file = rd.choice(psf_files)
```
HironaoMiyatake, line 102: I think it would be nice to add an option to specify a seed to make the choice of PSF reproducible.

ismael-mendoza: Thanks @HironaoMiyatake, I agree. This is related to #69 and it would be great to make this consistent throughout BTK to ensure galaxy blend + psf reproducibility. I would probably leave this for a separate PR that closes #69 but this is definitely important.

HironaoMiyatake: I agree. Let's close this PR and work on reproducibility in a PR for #69.

btk/draw_blends.py
mpaillassa: Yes indeed !

mpaillassa:
I just pushed the two fixes mentioned above and set the PSF static by default for Rubin survey (as for all other surveys).

About the tutorial, do you think it is better to explain this feature in a separate notebook using catsim or to do it in the galsim notebook (once the issue #91 is solved) ?

thuiop:
It probably goes in the intro notebook (I see it has been drastically reduced !) or something. Anyway, I think I will go along with @ismael-mendoza suggestion to integrate notebooks to the documentation (which seems possible) ; so we can leave it for now and I will take care to implement it when I get to work on the documentation (ie when this PR is closed).

thuiop (approved):
I think the PR is pretty much ready to be merged.

PR #105, opened 2021-03-01
Update documentation

### LINKED ISSUE #68: Documentation
The doc for BTK is largely outdated and incomplete ; it needs some working over.

thuiop:
I retrieved the content from my previous branch on documentation and have begun updating it to match the current version of BTK, to fix #68 . The autodoc should be working, though not all methods and function currently have a docstring ; one of the objectives of this PR should be to fix this in order to get a clean autodoc.
I am also investigating the integration of jupyter notebooks in the documentation, which is currently working on my computer.
Finally, we should modify the CI to build the documentation automatically, and will probably switch to Github Pages instead of readthedocs.


thuiop:
The autodoc part is mostly done, I ensured every relevant class and function had a proper docstring (except for the measurement/metrics part which will be reworked soon anyway) and that the result is as clean as possible. A few things to complete : if someone can give me a proper definition of zeropoint_airmass in the Filter object we can add it, and the GroupSamplingFunction parameters are not well described. 

What remains to be done in this PR are the main pages : 
- the front page can use some rewriting on the features part
- the "installation" page should be rewritten ; I think @ismael-mendoza is working on making BTK installable via pip, so I will let him write this page once it is done
- the "catalog" page is fine
- the user guide has already been updated, may need a few tweaks based on your opinions
- the tutorial page will be my next goal

The other thing to do is setup the CI for building the documentation automatically and plug it in with GitHub pages ; I would like to do this before the tutorial page to ensure that jupyter-sphinx works correctly.

thuiop:
@ismael-mendoza can you take care of the Github Pages stuff ? You should be more familiar than me with how it works. I think this can be useful for integrating directly with sphinx (which is used to build the documentation) https://github.com/ammaraskar/sphinx-action-test/blob/master/.github/workflows/default.yml

ismael-mendoza:
Thanks @thuiop for all your work on this. I just started to take a look and I'll setup the GitHub Pages

docs/source/tutorials.rst
```diff
@@ -16,26 +18,51 @@ The notebook also shows examples of performing:
...
-btk generates postage stamp images of blend scenes from an input catalog. How a particular scene is sampled from this catalog is defined in a function input to *btk.create_blend_generator.generate*.
-This `notebook <https://github.com/LSSTDESC/BlendingToolKit/blob/master/notebooks/custom_sampling_function.ipynb>`_ demonstrates how users can define their own sampling function to draw blend image scenes.
 
-Also shown is an example of how a user-defined custom function can be used to generate different observing conditions. This enables on-the-fly generation of blend scene images with different noise levels and observing PSFs.
+Tests
+------
+
+Importing relevant modules
+
+.. jupyter-execute::
```
ismael-mendoza, line 27: Interesting, are you needing to type the Jupyter notebook contents directly into this `.rst` file?

docs/source/src/btk.survey.rst
```diff
@@ -0,0 +1,12 @@
...
+The survey module defines the :class:`~btk.survey.Survey` class as a namedtuple, which contains the information relevant to observing conditions for a given survey. A Survey contains several instances of :class:`~btk.survey.Filter`, which are also namedtuples containing information specific to each filter. 
+
+.. automodule:: btk.survey
+	:exclude-members: Survey, Filter
+	
+	.. autoclass:: Survey
+		:exclude-members: airmass, filters, effective_area, mirror_diameter, zeropoint_airmass, pixel_scale, name, count, index
+
+	.. autoclass:: Filter
+		:exclude-members: exp_time, extinction, name, psf, sky_brightness, zeropoint, name, count, index
```
ismael-mendoza, line 12: So `Survey, Filter` appear in the autodoc? It seems you added `doc` strings for them

docs/source/conf.py
```diff
@@ -88,12 +88,21 @@ def __div__(self, other):
...
     "sphinx.ext.napoleon",
+    "jupyter_sphinx",
 ]
 
+autodoc_default_options = {
+    "members": True,
+    "special-members": "__init__",
+    "undoc-members": True,
+    "show-inheritance": True,
+    # "exclude-members": "__weakref__",
```
ismael-mendoza, line 103: you can remove this commented out stuff, although not sure what it does

.github/workflows/test.yml
```diff
@@ -54,6 +54,10 @@ jobs:
         run: |
           pipenv install -r dev-requirements.txt
 
+      - name: Install doc dependencies
+        run: |
+          pipenv install -r doc-requirements.txt
+
```
ismael-mendoza, line 60: it might make sense to have a separate workflow for building the docs (like `docs.yml` or something). Since it would be useful to know if say the tests are failing but the docs are building correctly in a given PR.

.gitignore
```diff
@@ -3,3 +3,6 @@
 /my-notebooks/
 .vscode/
 .pytest_cache/
+/build/
+.coverage
+/docs/build/
```
ismael-mendoza, line 8:
```suggestion
/docs/build/

```

ismael-mendoza: Please add newlines at the end of every file.

btk/catalog.py
```diff
@@ -70,7 +86,7 @@ def _prepare_table(self):
         # BTK now requires ref_mags, but Catsim still wants magnitudes
         table["ref_mag"] = self._raw_catalog["i_ab"]
         table["btk_size"] = r_sec
-        # Adds the extra columns to both catalogs just to be sure
+        # Adds the extra columns to both catalogs just to be sure - TO CHECK
```
ismael-mendoza, line 89: what needs to be checked exactly?

btk/catalog.py
```diff
@@ -8,15 +8,20 @@
 
 
 class Catalog(ABC):
+    """Base class containing the catalog for BTK.
```
ismael-mendoza, line 11: Thanks for adding all the docstrings :)

btk/survey.py
```diff
@@ -22,6 +22,18 @@
     ],
 )
 
+Survey.__doc__ = """
```
ismael-mendoza, line 25: Cool! I didn't know you could do this. Does it display the doc correctly in the auto doc?

ismael-mendoza (commented):
Thanks @thuiop , I had some questions and comments in this review. I will also help with the bullet points you listed after I setup the GitHub Pages.

docs/source/tutorials.rst
thuiop: Basically yes, the jupyter-sphinx library import those, create an actual notebook, executes it and put the results back in. It is not interactive though but I believe it is enough for the documentation.

thuiop: (I started rewriting an actual notebook and will transfer the content in the doc afterwards)

docs/source/src/btk.survey.rst
thuiop: Yes, they do ; I managed to add some documentation on it in the survey.py

.github/workflows/test.yml
thuiop: As you prefer ; it would make sense to have one.

btk/catalog.py
thuiop: More of reminder than actual checking ; I believe the two following lines are useless ; this will probably be tackled in the COSMOS PR though so adding the comment was not that useful

btk/survey.py
thuiop: Yes it does ; you can actually do this with other python objects, it is equivalent to putting the docstring just after the function signature.

ismael-mendoza:
Ok, so I created a separate workflow but the build is failing because `sphinx-rtd-theme` is not installed? Strangely it's in the `docs.requirements.txt` so this is mysterious to me.

thuiop:
> Ok, so I created a separate workflow but the build is failing because `sphinx-rtd-theme` is not installed? Strangely it's in the `docs.requirements.txt` so this is mysterious to me.

Strange as it works for me locally, but I'm not attached to it anyway (it only defines the style as the default readthedocs style) ; I will remove it and see if it works.

> @thuiop , I also setup the GitHub Pages and I think it should work as long as we correctly do the build i.e. we should have `.html` files sitting inside the `/docs` folder that get pushed to the `main` branch. Not exactly sure how this action works but hopefully that's part of what it does?

Great, thanks. I think the action mostly executes the Makefile in docs/ (with `make html`), which generates a docs/build folder ; the html files are in that one.

> Separately, the `pytest` is failing, but the error is also cryptic to me (related to `pipenv`) I will probably fix this in a different branch

I'm not sure why, I will take a looK.

docs/source/tutorials.rst
ismael-mendoza: sounds good, are you also able to use an existing notebook instead of having to rewrite it inside the `.rst` document?

ismael-mendoza:
> Strange as it works for me locally, but I'm not attached to it anyway (it only defines the style as the default readthedocs style) ; I will remove it and see if it works.

If all the action does is build the html and you have a way to this locally (what is the command you used?). You might also try an approach similar to [this](https://github.com/ammaraskar/sphinx-action-test/blob/f9cb44b1fdc634b1e7d71ff0721400226babd4ad/.github/workflows/default.yml#L43). We could even have a separate branch just for documentation `gh-pages` which seems more standard from what I could read (that way documentation is not build in every PR to the `main` branch).

I suspect the GitHub action has a separate sphinx package stuff it uses so that why the imports are failing. So if there is a way to use what is pip installed that seems easier.

ismael-mendoza:
> I'm not sure why, I will take a looK.

Thanks, I'm pretty sure it's not your fault actually (since I committed something super small to the main branch and it failed). So if you want to continue looking into the docs and I'll take care of this pytest stuff, this would be OK with me.

thuiop:
> > Strange as it works for me locally, but I'm not attached to it anyway (it only defines the style as the default readthedocs style) ; I will remove it and see if it works.
> 
> If all the action does is build the html and you have a way to this locally (what is the command you used?). You might also try an approach similar to [this](https://github.com/ammaraskar/sphinx-action-test/blob/f9cb44b1fdc634b1e7d71ff0721400226babd4ad/.github/workflows/default.yml#L43). We could even have a separate branch just for documentation `gh-pages` which seems more standard from what I could read (that way documentation is not build in every PR to the `main` branch).
> 
> I suspect the GitHub action has a separate sphinx package stuff it uses so that why the imports are failing. So if there is a way to use what is pip installed that seems easier.
> 
> There is also an option to run your own [build command](https://github.com/ammaraskar/sphinx-action-test/blob/f9cb44b1fdc634b1e7d71ff0721400226babd4ad/.github/workflows/default.yml#L22). But not sure if that makes a difference.

I agree, this may be better anyway. (I use make html locally)

thuiop:
Seems like we are getting the same error for docs that we have for pytest ; it seems to me to be github related (the env not being created properly ?). It might be related with the sphinx-action since it is basically the only thing which was modified in the pytest workflow

thuiop:
Well the workflow is back to exactly how it was before this PR but still fails...

thuiop:
@ismael-mendoza It seems like disabling the pipenv part solved the problem. You may want to take a look at it some time ; for now I will see if I can make the documentation work.

docs/source/tutorials.rst
thuiop: Well I have to rewrite it but it is pretty straightforward to do (mostly copy and paste). I think there is a way of just integrating a notebook inside the doc directly but I think it is better as it is now.

thuiop:
> > Strange as it works for me locally, but I'm not attached to it anyway (it only defines the style as the default readthedocs style) ; I will remove it and see if it works.
> 
> If all the action does is build the html and you have a way to this locally (what is the command you used?). You might also try an approach similar to [this](https://github.com/ammaraskar/sphinx-action-test/blob/f9cb44b1fdc634b1e7d71ff0721400226babd4ad/.github/workflows/default.yml#L43). We could even have a separate branch just for documentation `gh-pages` which seems more standard from what I could read (that way documentation is not build in every PR to the `main` branch).

OK the doc is now building properly, without using the action. I think we can try having a gh-pages branch as you mention ; I will do that in the afternoon.

thuiop:
Good news, the documentation is now building ! You can see the result at https://lsstdesc.org/BlendingToolKit/index.html. I created a gh-pages branch in this repo, to which the changes are pushed ; it is a bit ugly but it works and we should now leave that branch alone. Right now the documentation will update each time there is a push in a PR ; I think it is better to leave it that way until we close the PR, and then switch it to rebuild only when PR are merged.

ismael-mendoza (approved):
Looks pretty good @thuiop, should we merge or is there something else you wanted to do?

thuiop:
There are still a few things I want to change, I'll sort these up this weekend and we should be able to merge on Monday.

ismael-mendoza:
> There are still a few things I want to change, I'll sort these up this weekend and we should be able to merge on Monday.

Great! Side note: I noticed you added a `.nojekyll` object and I was wondering if using one of the themes might help make some of the auto docs even easier to parse? I'm also ok just having the simplest thing for now and adding themes later but I was just curious

thuiop:
My understanding was that the .nojekyll was necessary when using sphinx with GitHub actions otherwise they conflict with each other. One of the things I wanted to do was changing the theme within sphinx ; I might go for the same as astropy.
The main thing missing is the tutorial ; I have already written it down so I will include it in this PR. There are also a few outdated things remaining so I will also clean these.

thuiop:
This branch should be ready to be merged now, provided the tests pass and that everyone is satisfied with the documentation as it is now.

ismael-mendoza (approved):

PR #107, opened 2021-03-04
Cosmos-catalog

mpaillassa:
Following the discussion in issue #91, I start this PR to solve the issues regarding the COSMOS catalog (#91 and #80).

For now I have just added a galsim COSMOSCatalog attribute to the `CosmosCatalog` class. It can be retrieved in `render_single` to use the `makeGalaxy` method (it should be stored as an attribute instead of being retrieved at each `render_single` call).
In order to make the galaxy designed by `entry` with `makeGalaxy`, I assumed that both `CosmosCatalog` and `COSMOSCatalog` were read in the same order. But I am not sure that it is correct, and I don't really know why but the catalogs don't have the same number of entries.. (`CosmosCatalog` is larger, so it fails sometimes).

About the galaxy fluxes, I just used the `get_flux` function with the `ref_mag` from the `CosmosCatalog` but I am not sure it is the way to do it.


thuiop:
Isn't the size difference due to the fact that there are two files imported for the catalog ? Try removing the catalog_fit line ; does it make a difference ? Otherwise, the process seems fine to me ; if we have a separate variable I am not sure that we really need _raw_catalog anymore.

Regarding the fluxes, it could be as simple at that. Did you draw some example of blends and do the fluxes look right to you ? You can put some here if you want.

SIde note : you may want to use the pre-commit to pass the linting on the first try :)

btk/catalog.py
```diff
@@ -78,26 +79,33 @@ def _prepare_table(self):
...
 
     def _prepare_table(self):
         table = deepcopy(self._raw_catalog)
         # make notation for 'ra' and 'dec' standard across code.
         table.rename_column("RA", "ra")
         table.rename_column("DEC", "dec")
         table.rename_column("MAG", "ref_mag")
-        index = np.where(t["IDENT_1"] == self._raw_catalog["IDENT_1"] for t in table)
-
+        #index = np.where(t["IDENT_1"] == self._raw_catalog["IDENT_1"] for t in table)
```
ismael-mendoza, line 106: remove this commented out version?

btk/draw_blends.py
```diff
@@ -415,8 +415,10 @@ def render_single(self, entry, filt, psf, survey):
         Returns:
             gal (galsim.InterpolatedImage): The galsim profile of a single galaxy
         """
-        k = int(np.random.rand(1) * len(self.catalog))  # catalog_line["btk_index"][0]
-        gal = self.catalog.makeGalaxy(k, gal_type="real", noise_pad_size=0).withFlux(1)
+
+        galsim_catalog = self.catalog.get_galsim_catalog()
+        gal_flux = get_flux(entry['ref_mag'], filt, survey)
```
ismael-mendoza, line 420: This looks right to me, although I know there are several definitions of magnitudes floating around. Is this a AB magnitude? (I assume this is the standard convention for most things ? )

btk/catalog.py
```diff
@@ -109,3 +117,6 @@ def _prepare_table(self):
         self._raw_catalog["ref_mag"] = self._raw_catalog["MAG"]
         self._raw_catalog["btk_size"] = size
         return table
+
+    def get_galsim_catalog(self):
+        return self.galsim_catalog
```
ismael-mendoza, line 122: Yes, as @thuiop  mentioned. I think we should keep either raw_catalog or galsim_catalog but not both. I think `_raw_catalog` was a way to standardized the different catalogs. But since we only have two types I don't see it like a super useful attribute right now. I you want to remove it and just keep `galsim_catalog` that is fine.

btk/draw_blends.py
```diff
@@ -415,8 +415,10 @@ def render_single(self, entry, filt, psf, survey):
...
+        gal = galsim_catalog.makeGalaxy(entry['btk_index'], gal_type="real", noise_pad_size=0).withFlux(gal_flux)
```
ismael-mendoza, line 421: Is it important to set the `noise_pad_size` to something ? Yes, maybe looking at some images might be helpful as @thuiop suggested above. Feel free to paste a couple in the PR.

ismael-mendoza (changes requested):
Thanks @mpaillassa this is a good start! I had a couple of questions and comments below which also address some of your questions

btk/draw_blends.py
thuiop: I think this should be managed in the _prepare_table function in all cases, ref_mag is one of the BTK specific arguments. This should be the AB magnitude for the function to give the correct result though.

btk/catalog.py
thuiop: The only time where it would be useful would be if we had a DrawBlendsGenerator working for several different catalogs but I do not see that happening often.

btk/catalog.py
mpaillassa: > Isn't the size difference due to the fact that there are two files imported for the catalog ? Try removing the catalog_fit line ; does it make a difference ? Otherwise, the process seems fine to me ; if we have a separate variable I am not sure that we really need _raw_catalog anymore.

The size of `catalog_coord` and `catalog_fit` are the same (87798) while the galsim catalog has 81499 entries...
If we remove `_raw_catalog`, we would have to have it as a parameter of `_prepare_table`, isn't it ? But I think we should still read `catalog_coord` and `catalog_fit` at some point because I am not sure we can retrieve the necessary information in the galsim catalog.

> Regarding the fluxes, it could be as simple at that. Did you draw some example of blends and do the fluxes look right to you ? You can put some here if you want.

It seems ok but I think we should solve the problem of shifts to assess this. Currently shifts are not used. I guess we should use `entry['ra']` and `entry['dec']` in `render_single`. 
 
> SIde note : you may want to use the pre-commit to pass the linting on the first try :)

Indeed. I just switched from a fork to a clone and forgot to install it :)

mpaillassa:
> Isn't the size difference due to the fact that there are two files imported for the catalog ? Try removing the catalog_fit line ; does it make a difference ? Otherwise, the process seems fine to me ; if we have a separate variable I am not sure that we really need _raw_catalog anymore.

The size of `catalog_coord` and `catalog_fit` are the same (87798) while the galsim catalog has 81499 entries...
If we remove `_raw_catalog`, we would have to have it as a parameter of `_prepare_table`, isn't it ? But I think we should still read `catalog_coord` and `catalog_fit` at some point because I am not sure we can retrieve the necessary information in the galsim catalog.

> Regarding the fluxes, it could be as simple at that. Did you draw some example of blends and do the fluxes look right to you ? You can put some here if you want.

It seems ok but I think we should solve the problem of shifts to assess this. Currently shifts are not used. I guess we should use `entry['ra']` and `entry['dec']` in `render_single`. 
 
> SIde note : you may want to use the pre-commit to pass the linting on the first try :)

Indeed. I just switched from a fork to a clone and forgot to install it :)

thuiop:
I am not sure about this catalog length thing then. For the _raw_catalog, I mean we don't need to store it ; we could give it as an argument for _prepare_table for instance.
For the shifts, you can use gal.shift(entry["ra"], entry["dec"]) just like for the Catsim one.

btk/draw_blends.py
ismael-mendoza: do you know if it's an AB magnitude? @thuiop or @mpaillassa

btk/draw_blends.py
mpaillassa: I don't know I guess we don't want noise in galaxy images as we add the Poisson noise later in the simulation ?

mpaillassa:
I pushed some changes to remove the `_raw_catalog` attribute and use the position shifts.

About the difference length of `CosmosCatalog` and `COSMOSCatalog`, I found that galsim automatically applies a [cut](https://galsim-developers.github.io/GalSim/_build/html/real_gal.html#galsim.COSMOSCatalog) with `exclusion_level` to eliminate "bad" galaxies. For now I just disabled it. If we want to apply it we should find a way to make the same cut on the raw catalog that we read to build `table`.

About the fluxes, we still have to check if these are calculated with the right magnitude (AB).

thuiop:
There is something I did not mention yesterday because I was not sure about it on the flux issue. In the code we used before switching to BTK, we had some real galaxies drawn in multiband ; I looked into it, and it turns out there may be another approach for the fluxes. Basically, when drawing parametric galaxies from COSMOS, there is a chromatic option which can be activated for multiband images ; it makes some basic assumptions regarding the SEDs of the disk and the bulge, and it allows for drawing the object with different filters by giving a galsim.Bandpass object, which contains information about the filter (the ones we have are built from files containing the throughput of the filter on many wavelength). This only works for parametric though, but what we do is rescale the flux of the real image with the flux of the image in the band we want. This may be an interesting way to solve the problem, what do you think @herjy ?

mpaillassa:
If it is the solution that we adopt, I have seen in the galsim doc the `chromatic` option of `makeGalaxy` but I don't see how to incorporate a given band information via a `bandpass` object. Should it be given as a `gsparams` argument ?

thuiop:
> If it is the solution that we adopt, I have seen in the galsim doc the `chromatic` option of `makeGalaxy` but I don't see how to incorporate a given band information via a `bandpass` object. Should it be given as a `gsparams` argument ?

If your question is about how to create the bandpass, we have been using files containing the informations about the filter to create the bandpass ; this is one of the main drawbacks since it means that you need such files to be able to use this technique (though I think there are other ways of doing it). Here is an example of such file so that you get an idea https://raw.githubusercontent.com/lsst/throughputs/master/baseline/total_u.dat .

If you question was about what to do with the bandpass once you have it, you use it in the drawImage method, eg `drawImage(bandpass,image=image)` instead of `drawImage(image=image)`.

mpaillassa:
Ok thank you ! So the plan is to go for it if @herjy thinks it's ok ? 
If so we would need to retrieve the bandpass files for each filter.

thuiop:
Well it was just an idea, it sounds more convincing to me than getting the fluxes from Catsim but I'm not an expert, I could be wrong. Anyway I should be able to provide the files for LSST, a bunch of HST filters and Euclid.

ismael-mendoza:
Just to offer my two cents, the color issue seems a bit complicated and (perhaps?) out of the scope of this PR. I would be happy to get the COSMOS galaxies working in 1 band and then have another PR later to make it work on multiple colors. 

Of course, this is up to you @mpaillassa, but on my end it would be nice see this working on 1 band in some consistent way (maybe with an accompanying notebook?). So at least we have some basic stuff running. 

In terms of the multi-band stuff, what @thuiop suggests sounds reasonable to me. It makes more sense than using the CATSIM colors. But I'm also not too familiar with this so I think asking an expert would be the best at this point. Maybe we can ping someone in the LSST slack that developed this feature in galsim? (e.g. Rachel or Mike?)

HironaoMiyatake:
I also think working on 1 band for this PR makes sense. We can open a new PR for implementing multi-band image generation. So the galsim multi-band feature works only for parametric galaxies, right? If so, we still think about a workaround for real images.

thuiop:
> I also think working on 1 band for this PR makes sense. We can open a new PR for implementing multi-band image generation. So the galsim multi-band feature works only for parametric galaxies, right? If so, we still think about a workaround for real images.

Yeah, basically Galsim only natively support multiband for the parametric galaxies. The proposition was to use the flux from the parametric model to rescale the flux of the real galaxy for each band.

HironaoMiyatake:
How do they compute fluxes for a parametric galaxy? They use the bulge/disk fraction? If so, do we have a similar quantity for real galaxies (maybe in a real galaxy catalog if it exists)?

thuiop:
> How do they compute fluxes for a parametric galaxy? They use the bulge/disk fraction? If so, do we have a similar quantity for real galaxies (maybe in a real galaxy catalog if it exists)?

The doc of galsim is not very clear on this point ; my understanding is that they assume an SED for the bulge and one for the disk and work from there ; I do not know the exact details, I will check in the code.

HironaoMiyatake:
Thank you for looking into this! So if there is a bulge/disk fraction in the real galaxy catalog, it's straightforward to implement chromaticity.

Anyway, I am happy for this PR to be merged before adding this feature. I can make another PR later.

mpaillassa:
Following the Slack discussions, I guess we can think about merging this PR and going for another issue/PR for multiband support with Cosmos galaxies.
Should we merge this PR in the current state ? Currently it uses the same flux for every band, computed from the AB magnitude from the galsim COSMOSCatalog.

ismael-mendoza (dismissed):
Yes I think we can merge, @mpaillassa  if possible could you add a smoke unit test to show how use this catalog/class? Feel free to add the COSMOS catalog data files to the `/data` folder via [git-lfs](https://git-lfs.github.com) if that sounds good to you.  

You might also need to rebase the branch, let me know if this causes any problems.

mpaillassa:
I added the cosmos catalogs with git-lfs and some reading tests. Are these tests ok ?

ismael-mendoza (approved):
Thanks @mpaillassa for your work on this, I think we can merge now :)

PR #112, opened 2021-03-08
Adding new tests to improve coverage

thuiop:
I noticed that test coverage is pretty low right now, so I am adding some tests to improve it, mainly by testing the cases where there should be an error.

### DIFF
--- a/tests/test_error_cases.py
+++ b/tests/test_error_cases.py
@@ -1,0 +1,150 @@
+import pytest
+
+import btk.sampling_functions
+from btk.survey import Rubin
+
+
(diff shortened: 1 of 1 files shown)


thuiop:
You can merge whenever, I think I'm finished for this PR. I mostly added tests that just run the code and see if there are not any errors and tests that check some error cases. We should add tests for the COSMOS image generation when it will finally work, and we will have to rework the tests for measure and metrics anyway.

ismael-mendoza (approved):
Thanks @thuiop , I learned more about `pytest` from reviewing your PR :). I will merge now

PR #113, opened 2021-03-11
switch band order to scarlet band order

ismael-mendoza:
This is more of a selfish change, I've always wanted the order of the images shape to be [n_bands, nx, ny] rather than [nx, ny, n_bands]. This is also the convention used in scarlet and in [bliss](https://github.com/prob-ml/bliss) so it would be nice to have everything consistent and not have to remember to transpose (at least for me). This also simplifies the code in `plot_utils.py` slightly. 

Would this cause any problems for anyone? Did I miss any places?


thuiop:
Well I do usually have the channels as the last dimension... maybe we should have a tag defining the order and doing the transposition automatically ?

ismael-mendoza:
> Well I do usually have the channels as the last dimension... maybe we should have a tag defining the order and doing the transposition automatically ?

Yeah I can do that, I think I will use this order as the default for everything, but add an optional keyword argument to the draw_blend_generator so that it can return the images transposed in a specific way. Would that work for you?

thuiop:
I think it works that way. Tensorflow uses a string with values "NHWC" or "NCHW" as a tag (for channel last and channel first), this may be more user-friendly ? I don't think anyone would want to have the channels as the middle one anyway. But it's a very minor thing anyway.

thuiop (approved):
See my last comment ; apart from that it seems OK for me, as long as the tests run it should be working.

ismael-mendoza:
> I think it works that way. Tensorflow uses a string with values "NHWC" or "NCHW" as a tag (for channel last and channel first), this may be more user-friendly ? I don't think anyone would want to have the channels as the middle one anyway. But it's a very minor thing anyway.

Ah I like this idea better actually, I will quickly make the change and then merge when tests are passing. Thanks!

Issue #100, opened 2021-02-25
Reworking the measurement and metrics part


thuiop:
During the last months, we have mainly focused on image generation which is only one part of BTK (albeit an important one). I believe the next big pull request should be about the measurement part which have not been updated for a while (but should still be functional). 
I'm putting here the up-to-date flowchart of BTK to show how it works.
![btk(4)](https://user-images.githubusercontent.com/1338337/109148692-e0fd5380-7766-11eb-937b-b3b8455e56db.png)
The user is supposed to create a subclass of MeasurementParams, with custom get_deblended_images and make_measurement methods (though make_measurement is basically unused in the default implementation), which should return respectively the deblended images and the measurements on the blended images. They give this to the MeasureGenerator, as well as a DrawBlendsGenerator, which when called generates the images and executes the two methods on them (and returns all of it).
For the metrics (actually measuring the performance of the algorithm), they should create a subclass of MetricsParams, several function such as get_detections, get_segmentation, get_flux... which are used to recover the results from the MeasureGenerator (which is given to the MetricsParams). The MetricsParams is then given to the run function, which uses a bunch of utility functions to evaluate the performance.
As you can see, it lacks a bit of symmetry. I think this part definitely need some reworking to get something people can actually and understand and use. Please put your ideas under this issue if you have any.

What I think : the generator structure seems appropriate to me ; I am ok with giving the DrawBlendsGenerator to the MeasureGenerator. I believe there should also be a MetricsGenerator receiving the MeasureGenerator. The MeasurementParams seems a bit heavy to me ; basically it's there to gather several functions made by the users to run on the generated data. I think this could be achieved using a namedtuple (to keep consistency with other parts of BTK), with several "special" attributes corresponding to specific measurements which will be used for the metrics part (eg deblended images, detections, segmentation,...), where users would provide functions they made, and possibly an additionnal attribute if users want to run some other measurements on the data not covered by the metrics part. EDIT : I realized that this would not be a very good way to do it, as often the same algorithm gives both segmentation, deblended images, centers or whatever. Instead, maybe the user should provide an unique function returning the results as a dictionnary with specific keys ?
In the same fashion, the MetricsParams is too heavy, and actually I do not think it really is useful (the fact that it takes in a MeasureGenerator is also very weird.
Anyway, there is some work to be done to sort out the metrics part as there are a lot of functions in it right now, and it is unclear to me right now which are useful and which are not.

thuiop:
[...]

PR #114, opened 2021-03-11
Rework measurement part

ismael-mendoza:
This simplifies the measurement part of BTK taking into account suggestions in #100. Documentation still needs to be updated accordingly, but it would get to get some feedback before finalizing that. Thanks! 

@thuiop let me know if you have any additional suggestions. 

NOTE:  The diff is a little hard to understand so maybe looking at the file directly is better.


btk/measure.py
```diff
@@ -1,119 +1,73 @@
...
-        return None
+    Args:
+        batch (dict): Output of DrawBlendsGenerator object's `__next__` method.
+        idx (int): Index number of blend scene in the batch to preform
+            measurement on.
 
-    def get_deblended_images(self, data, index):
-        """Function describing how the deblending algorithm is run.
+    Returns:
+            dict containing subset of keys from: ['deblend_image', 'peaks']
```
thuiop, line 32: I would rename the key in "deblended_images" for consistance. We'll also need to specify which format we want for all of the entries somewhere. Finally, we may want additionnal entries ? I'm thinking about segmentation in particular. Though we can probably handle in the metrics part the case where the key does not exist.

btk/measure.py
```diff
@@ -1,119 +1,73 @@
...
-        Returns:
-            dict with the centers of sources detected by SEP detection
-            algorithm.
-        """
-        image = np.mean(data["blend_images"][index], axis=2)
-        peaks = self.get_centers(image)
-        return {"deblend_image": None, "peaks": peaks}
+    image = np.mean(batch["blend_images"][idx], axis=2)
+    bkg = sep.Background(image)
+    catalog = sep.extract(image, 1.5, err=bkg.globalrms, segmentation_map=False)
```
thuiop, line 62: Here we can easily get the segmentation with no additionnal effort, and use it to get some basic deblended images ; I believe it will be useful both as an example and as a way to test the metrics part later. I can do it if you prefer.

btk/measure.py
```diff
@@ -143,33 +108,35 @@ def __init__(
...
-    def run_batch(self, blend_output, index):
-        deblend_results = self.measurement_params.get_deblended_images(
-            data=blend_output, index=index
-        )
-        measured_results = self.measurement_params.make_measurement(data=blend_output, index=index)
-        return [deblend_results, measured_results]
+    def run_batch(self, batch, index):
+        output = []
+        for f in self.measure_functions:
+            output.append(f(batch, index))
```
thuiop, line 114: I am unsure about which format to use exactly for the function. Shouldn't we only give it batch['blend_images'][index] ? Maybe adding the psf ? This does work so either way may be fine.

thuiop (commented):
Great job ; I made some minor comments. We should probably get the opinion from someone else before merging (pro tip : go look directly at the file instead of the diff here, it will be much clearer)

btk/measure.py
ismael-mendoza: I think as long as we are clear what the batch contains it should be ok for the user. I'm just giving the user all the info assuming that there might be different algorithms that require different info

btk/measure.py
ismael-mendoza: Yeah do you have suggestions for specifying the format? I added some stuff in the top of the file but not sure if that is the best place. I could also add assertions checking the output of measurement in the `run_batch` method.

btk/measure.py
ismael-mendoza: I just added it, let me know if that makes sense (I had removed it and was planning on adding it back later anyways)

ismael-mendoza: This got me thinking, maybe we want to enforce that every algorithm returns a `peak` key because otherwise there is no way of telling how to match the `deblended_images` or `segmentation` to which isolated objects.

btk/measure.py
thuiop: I think it should show up in the doc if it is at the top of the file, so it should be alright. Adding checks may help users troubleshooting their problems, this is a good idea.

btk/measure.py
thuiop: I think you can match even if you don't have the peaks, using intersection over union or something like that. It is easier to do with the peaks though.
The catalog idea is pretty good ; we should still make sure that people use standardized names but I think it would work great if we want to have support for photometric redshifts or shape measurement for instance.

thuiop: As for SEP, we can even have deblended_images by using the mask from the segmentation to recover the pixels corresponding to each object. I also think the raw segmentation from sep isn't a good format for segmentation in general as you can only have one object corresponding to each pixel ; we should instead have a (stamp_size,stamp_size,n_objects) boolean array.

ismael-mendoza: Thanks for reworking the SEP part, I agree with your reasoning. I'll try to add a catalog format with standard names.

ismael-mendoza:
Hi @thuiop, I think this is ready for another review if you have sometime. Then we can discuss what we might be missing in terms of documentation/tutorials or if we want to do that separately. 

Once finalized, I'll ping Alexandre again to see if he can give us a final review.

btk/measure.py
```diff
@@ -1,175 +1,234 @@
...
+                                     `len` of the table should be `n_objects`. Currently the
+                                     following column names are supported:
+                                        - dx: horizontal centroid position in pixels.
+                                        - dy: vertical centroid positoin in pixels.
+                                        - ra: object centroid right ascension in arcseconds,
+                                        following the convention from the `wcs` object included in
+                                        the input batch.
+                                        - dec: vertical centroid position in arcseconds,
+                                        following the convention from the `wcs` object included in
+                                        the input batch.
```
thuiop, line 17: You should specify that dx and dy should be used when there is a single survey and ra and dec when you have multiresolution.

btk/measure.py
```diff
@@ -1,175 +1,234 @@
...
+    catalog = astropy.table.Table()
+    catalog["dx"] = coordinates[:, 1]
+    catalog["dy"] = coordinates[:, 0]
+    return {"catalog": catalog}
+
+
+def sep_measure(batch, idx):
+    """Return centers detected when object detection and photometry
+    is done on input image with SEP.
+
```
thuiop, line 73: We may want to specify that this function only works when you have a single survey.

btk/measure.py
thuiop: Also, users should feel free to integrate any measurements they want in the catalog field, or even as other keys in the dictionnary ; those keys are mostly here to ensure we have a standardized format in the metrics, but they can have other measurements not covered by the metrics if they want.

thuiop (approved):
For me, this PR can be merged ; I have made a few small comments about the doc but overall it is ok for me.

thuiop (dismissed):

ismael-mendoza:
Waiting on a review from @aboucaud before merging. He said he can do it early next week, in the meantime I can start a branch with the metrics part based off of this branch.

btk/survey.py
```diff
@@ -527,15 +527,16 @@ def get_mean_sky_level(survey, filt):
...
+    that the center of the image in pixels [(s + 1) / 2, (s + 1) / 2] corresponds to
+    (ra, dec) = [0, 0].
 
     Args:
         pixel_scale (float): pixel size in arcseconds
         shape (tuple): shape of the image in pixels.
         center_pix (tuple) : tuple representing the center of the image in pixels
         center_sky (tuple) : tuple representing the center of the image in sky coordinates
                      (RA,DEC) in arcseconds.
-        projection(str): projection type, default to TAN. A list of available
```
aboucaud, line 538: Why has the projection been removed from the docstring ?

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+        idx (int): Index number of blend scene in the batch to preform
+            measurement on.
+
+    Returns:
+            dict containing catalog with entries corresponding to measured peaks.
+    """
+    if isinstance(batch["blend_images"], dict):
+        raise NotImplementedError("This function does not support the multi-resolution feature.")
+
+    image = np.mean(batch["blend_images"][idx], axis=0)
```
aboucaud, line 60: should be called coadd to be consistent with `sep_measure`.

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+
+    # set detection threshold to 5 times std of image
+    threshold = 5 * np.std(image)
+    coordinates = peak_local_max(image, min_distance=2, threshold_abs=threshold)
+
+    # construct catalog from measurement.
+    catalog = astropy.table.Table()
+    catalog["dx"] = coordinates[:, 1]
+    catalog["dy"] = coordinates[:, 0]
```
aboucaud, line 69: Not very important but the naming `dx` and `dy` for centres is a bit puzzling.
Naturally I would expect something like `x_cen` or `x_peak`.

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+            measurement on.
+
+    Returns:
+        dict with the centers of sources detected by SEP detection algorithm.
+    """
+    if isinstance(batch["blend_images"], dict):
+        raise NotImplementedError("This function does not support the multi-resolution feature.")
+
+    image = batch["blend_images"][idx]
+    stamp_size = image.shape[-2]  # true for both 'NCHW' or 'NHWC' formats.
```
aboucaud, line 91: This needs a longer comment, since it can lead to errors down the line. 
One should probably add in the method docstring that the blend images are ndarray expected in a certain format.

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+    coadd = np.mean(image, axis=0)
+    bkg = sep.Background(coadd)
+    catalog, segmentation = sep.extract(coadd, 1.5, err=bkg.globalrms, segmentation_map=True)
```
aboucaud, line 94: Where does the 1.5 come from ?
Since it's hardcoded, it should at least be commented.

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+    n_objects = len(catalog)
+    segmentation_exp = np.zeros((n_objects, stamp_size, stamp_size))
```
aboucaud, line 96:
```suggestion
    segmentation_exp = np.zeros((n_objects, stamp_size, stamp_size), dtype=bool)
```

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+    deblended_images = np.zeros((n_objects, *image.shape))
```
aboucaud, line 97:
```suggestion
    deblended_images = np.zeros((n_objects, *image.shape), dtype=image.dtype)
```

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+    for i in range(n_objects):
+        segmentation_exp[i][np.where(segmentation == i + 1)] = True
```
aboucaud, line 99: is the `np.where()` needed here ?
Just a gut comment, haven't checked.

btk/measure.py
thuiop: It's the threshold for detection against noise ; here it is set at 1.5 sigma. I agree that we should put up a comment.

btk/measure.py
thuiop: Now that I think of it, probably not.

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
-                                `btk.measure.Measurement_params`.
+            measure_function: Function or list of functions that returns a dict with
+                              measurements given output from the draw_blend_generator.
             draw_blend_generator: Generator that outputs dict with blended images,
-                                  isolated images, observing conditions and blend
-                                  catalog.
+                                  isolated images, blend catalog, wcs info, and psf.
             multiprocessing: If true performs multiprocessing of measurement.
             cpus: If multiprocessing is True, then number of parallel processes to
                  run [Default :1].
```
aboucaud, line 141: Not in this PR, but I think these two arguments are redundant.
```python
multiprocessing = True if cpu > 1 else False
```

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
-        measured_results = self.measurement_params.make_measurement(data=blend_output, index=index)
-        return [deblend_results, measured_results]
+    def run_batch(self, batch, index):
+        output = []
+        for f in self.measure_functions:
+
+            out = f(batch, index)
+
+            # make sure output is in the correct format.
+            if not isinstance(out["catalog"], astropy.table.table.Table):
```
aboucaud, line 175:
```suggestion
            if not isinstance(out["catalog"], astropy.table.Table):
```

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+                                    number of detected objects
+    - segmentation (np.ndarray): Array of booleans with shape `(n_objects,stamp_size,stamp_size)`
+                                 The pixels set to True in the i-th channel correspond to the i-th
+                                 object. The order should correspond to the order in the returned
+                                 `catalog`.
+
+Omitted keys in the returned dictionary are automatically assigned a `None` value (except for
+`catalog` which is a mandatory entry).
+"""
+import astropy
```
aboucaud, line 35:
```suggestion
import astropy.table
```
It seems ones now needs to import the submodule `astropy.table` in order to access the Table, like `scipy`, at least on my computer. I'm running astropy==4.2.

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
+                                len(measure_functions) corresponding to the measurements made by
+                                each function on each element of the batch. If only one
+                                measure_function was passed in, then each sub-list is converted
+                                to a single element.
         """
 
         blend_output = next(self.draw_blend_generator)
-        deblend_results = {}
-        measured_results = {}
         input_args = [(blend_output, i) for i in range(self.batch_size)]
```
aboucaud, line 226: I guess when you write such list comprehension, `blend_output` is copied multiple times, which is quite inefficient. 
Instead, a generator would be much better.
```suggestion
        input_args = ((blend_output, i) for i in range(self.batch_size))
```

btk/measure.py
```diff
@@ -1,175 +1,239 @@
...
             self.cpus,
             self.multiprocessing,
             self.verbose,
         )
-        for i in range(self.batch_size):
-            deblend_results.update({i: batch_results[i][0]})
-            measured_results.update({i: batch_results[i][1]})
+        for i, r in enumerate(measure_results):
+            if len(r) == 1:
+                measure_results[i] = r[0]
```
aboucaud, line 236: What is the purpose of this ?

tests/test_measure.py
```diff
@@ -44,10 +44,10 @@ def get_meas_generator(meas_params, multiprocessing=False, cpus=1):
 
 def compare_sep():
     """Test detection with sep"""
-    meas_param = btk.measure.SepParams()
-    meas_generator = get_meas_generator(meas_param)
-    _, deb, _ = next(meas_generator)
-    detected_centers = deb[0]["peaks"]
+    meas_generator = get_meas_generator(btk.measure.sep_measure)
+    _, results = next(meas_generator)
+    dx, dy = (results[0]["catalog"]["dx"].item(), results[0]["catalog"]["dy"].item())
```
aboucaud, line 49: IMO, one should be able to extract two result values with much straightforward code.
This would mean refactoring a bit the output of the measurement methods.
Probably for a next PR.

btk/measure.py
thuiop: We have a "list in, list out" approach for measure_functions ; if you provide a single function, you want to get your results directly, not as a list with a single element.

btk/survey.py
thuiop: Must be a mistake since next line is not removed.

btk/measure.py
thuiop: This should not be necessary since the images are supposed to come from a DrawBlendsGenerator directly, meaning they will be in the right format. When we make a tutorial for this, we should specify that images may come in the two different formats if users want to add their own measure_function.

btk/measure.py
```diff
@@ -91,12 +91,13 @@ def sep_measure(batch, idx):
...
+    # Here the 1.5 value corresponds to a 1.5 sigma threshold for detection against noise.
     catalog, segmentation = sep.extract(coadd, 1.5, err=bkg.globalrms, segmentation_map=True)
     n_objects = len(catalog)
     segmentation_exp = np.zeros((n_objects, stamp_size, stamp_size), dtype=bool)
     deblended_images = np.zeros((n_objects, *image.shape), dtype=image.dtype)
     for i in range(n_objects):
-        segmentation_exp[i][np.where(segmentation == i + 1)] = True
+        segmentation_exp[i][segmentation == i + 1] = True
         seg_i = (segmentation == i + 1).reshape(1, stamp_size, stamp_size)
         deblended_images[i] = seg_i * image
```
aboucaud: I suggest something even more straightforward
```suggestion
        seg_i = segmentation == i + 1
        segmentation_exp[i] = seg_i
        deblended_images[i] = image * seg_i[np.newaxis, ...]
```

btk/measure.py
aboucaud: Then it's no more "list in, list out". And it makes subsequent tasks more difficult because you may need to handle cases where the measurement is not a list..

btk/measure.py
ismael-mendoza: Thanks @aboucaud , I created a new issue to address this.

btk/measure.py
ismael-mendoza: I think @aboucaud is right, handling this special case might be problematic. I wonder if that's an issue with the multiple survey code as well

btk/measure.py
ismael-mendoza: Thanks I like this approach @aboucaud

btk/measure.py
ismael-mendoza: Yeah I think it can be puzzling too. I like your names better.

ismael-mendoza:
Thanks everyone for your help ! I'm merging this now

PR #125, opened 2021-04-07
rework metrics

ismael-mendoza:


thuiop:
Some thoughts about the structure : 
- For "basic" metrics, I think we only need a function that takes the information from a single blend (both the simulation and the measurements) and returns the corresponding metric ; some more complicated metrics may need to be applied on batches but we can think about this later
- We could have a big function that executes all these individual functions on the data. To make it compatible with results from outside BTK, this function should take a simplified version of the results with only the necessary information for computing the metrics.
- Then we can have a wrapper function for full BTK users, or maybe a generator to keep the same structure, which would convert the results from the measure generator to the right format and feed it to the previous function.
Does that seem right to you ? Do we need classes instead of functions at some point ?

thuiop:
I went ahead and wrote a tentative version of the metrics as I described ; feel free to comment on it ; again, you may want to look at the file directly and not at the diff. This is still largely incomplete but the structure is there.
Note that I also reorganized the output of the meas generator to be like `meas_results["sep_measure"]["catalog"][n_batch]`.
Feel free to test the code using the intro notebook.

ismael-mendoza:
Thanks @thuiop for your help on this, I'll post some review comments shortly

btk/measure.py
```diff
@@ -224,12 +224,21 @@ def __next__(self):
...
             self.cpus,
             self.verbose,
         )
         if self.verbose:
             print("Measurement performed on batch")
+        measure_results = {}
+        for i, f in enumerate(self.measure_functions):
+            measure_dic = {}
+            for key in ["catalog", "deblended_images", "segmentation"]:
+                if key in measure_output[0][i] and measure_output[0][i][key] is not None:
```
ismael-mendoza, line 239: I think you only need the second check in this 'and' clause (please let me know if I'm forgetting something though!)

btk/metrics.py
```diff
@@ -1,484 +1,110 @@
...
-    detected_table["match_galtileid1"] = np.ones(num_det, dtype=int) * -1
-    detected_table["match_galtileid2"] = np.ones(num_det, dtype=int) * -1
+import skimage.metrics
 
 
 def get_detection_match(true_table, detected_table):
     """Match detections to true objects and update values in the input
     blend catalog and detection catalog.
-
     Function does not return anything, only the astropy tables are updated.
```
ismael-mendoza, line 9: Is this a deprecated comment? Because it seems you are returning `match_table` and then using later in the code.

btk/metrics.py
```diff
@@ -1,484 +1,110 @@
...
     dist = np.hypot(t_x, t_y)
-    norm_size = true_table["size"]
-    norm_dist = dist / norm_size[:, np.newaxis]
-    detected_table["dSigma_min"] = np.min(norm_dist, axis=0)
     detected_table["d_min"] = np.min(dist, axis=0)
-    detection_threshold1 = 5
-    condlist1 = [
-        np.min(dist, axis=0) <= detection_threshold1,
-        np.min(dist, axis=0) > detection_threshold1,
+    detection_threshold = 5
```
ismael-mendoza, line 24: We probably want to make this changeable by the user.  Maybe even allow it to be a function in case people want to do things like have it in units of the size of the galaxy

btk/metrics.py
```diff
@@ -1,484 +1,110 @@
...
-    choicelist1 = [np.argmin(dist, axis=0), -1]
-    match_id1 = np.select(condlist1, choicelist1)
-    detected_table["match_true_id1"] = match_id1
-    detected_table["match_galtileid1"] = true_table["galtileid"][match_id1]
-    detection_threshold2 = 0.5
-    condlist2 = [
-        np.min(norm_dist, axis=0) <= detection_threshold2,
-        np.min(norm_dist, axis=0) > detection_threshold2,
+    choicelist = [np.argmin(dist, axis=0), -1]
+    match_id = np.select(condlist, choicelist)
```
ismael-mendoza, line 30: I actually never used this function before, but I think it works great in this case

btk/metrics.py
```diff
@@ -1,484 +1,110 @@
...
+    match_table["match_true_id"] = match_id
+    match_table["match_galtileid"] = true_table["galtileid"][match_id]
```
ismael-mendoza, line 32: Just to confirm I understand correctly, the `i`th element of `match_table['match_true_id']` is the index of the galaxy in `true_table` that is matched with the `i`th galaxy in `detected_table` right?

btk/metrics.py
```diff
@@ -1,484 +1,110 @@
...
-        # Evaluate flux measurement algorithm
-        flux = metrics_params.get_flux()
-        results["flux"].append(evaluate_flux(flux, index=i))
-        # Evaluate shape measurement algorithm
-        shapes = metrics_params.get_shapes()
-        results["shapes"].append(evaluate_shapes(shapes, index=i))
-    return results
+def compute_metrics_wrap(
+    blend_results, measure_results, use_metrics=("detection", "segmentation", "reconstruction")
+):
```
ismael-mendoza, line 101: This made me think about something @thuiop in order to simplify the UI - maybe we can keep the structure of a generator for the metrics so introduce a `MetricsGenerator` that processes one batch from the `MeasureGenerator` at a time (I think it's effectively what you have so far in the code?). 

I feel like from the user perspective, they probably don't want things in batches for analyzing the results after everything is processed. So in addition, we can provide an utility function that takes in a `metrics_generator` which processes some specified # of batches, calculates metrics for all of them, and summarizes the results into 3 simple tables. I can imagine the 3 tables being: 

- `true_table`: Contains true properties from all galaxies, in all blends. Each row is indexed by 'blend_id' and 'gal_id' (the index of the blend, and the index of each galaxy in that blend). 

- `measure_table`: Same as `true_table` but containing `match_id` (-1 if not matched) and only the info implemented in the `measure_function` via the `catalog` key. 

- `metrics_table`: Each row is indexed by `blend_id` and contains the metrics on that blend (# gals in that blend, # matches, mse, etc.) 

Some more global metrics that need to be accumulated independently can be returned additionally as a dictionary. What do you think @thuiop ?

ismael-mendoza:
Also not sure how many utility functions we need to provide for users who don't want to use a `measure_generator` (we ca also worry about this later), maybe each of the individual functions that takes in info from a given blend and returns the metric for that blend is enough?

btk/metrics.py
thuiop: Absolutely, I actually reused this function from the old metrics file and did not change the docstring yet.

btk/measure.py
thuiop: I think that if we try to access it and the entry does not exist it will throw an error. But I guess that they should be set to None anyway if they don't exist so you may be right.

btk/metrics.py
thuiop: Yes, we definitely want more flexibility in how the matches are done in the future, will get to it later.

btk/metrics.py
thuiop: In this version yes, but I changed it to be the other way around (we store the ids of the detected galaxies corresponding to each true galaxy).

btk/metrics.py
thuiop: We definitely need to think about the exact input and output format. Here my goal would be for users to have the data in the right format for the compute_metrics function and use it directly ; BTK users can use the "wrap" one directly, or we may have a generator which does that.
As for the output, we will definitely need some utility functions to generate nice plots and get summary statistics ; I feel like we can provide the "raw" metrics for every measure we actually make and either provide utility functions that do the analysis of these results, or even incorporate those directly into the process and directly return both the raw data and the summary data.

thuiop:
> Also not sure how many utility functions we need to provide for users who don't want to use a `measure_generator` (we can also worry about this later), maybe each of the individual functions that takes in info from a given blend and returns the metric for that blend is enough?
That's an important point and I have to think more about it.

btk/measure.py
ismael-mendoza: Yeah I always set them to `None`, I have it in the docstring too: ```Omitted keys in the returned dictionary are automatically assigned a `None` value (except for
`catalog` which is a mandatory entry).
```

Comment (ismael-mendoza) [reply btk/measure.py:239]: I think it's best to avoid checks that will never fail?

File: btk/metrics.py (hunk 3 above)
Comment (ismael-mendoza) [reply btk/metrics.py:27]: sounds good, maybe users can provide a function that calculates the distance based on the values of `x_peak` and `y_peak`, need to think more about how exactly

File: btk/metrics.py [hunk 7]
```diff
@@ -1,484 +1,167 @@
...
-        np.min(norm_dist, axis=0) > detection_threshold2,
+    choicelist = [np.argmin(dist, axis=0), -1]
+    match_id = np.select(condlist, choicelist)
+    # Not perfect... See how to improve that
+    for m in np.unique(match_id[match_id >= 0]):
+        idx = np.where(match_id == m)[0]
+        best_match = match_table["d_min"][idx].argmin()
+        for i in range(len(idx)):
+            if i != best_match:
+                match_id[idx[i]] = -1
```
ismael-mendoza, line 40: is this block of code because sometimes you could have multiple matches in `match_id` to the same true object?

btk/metrics.py
```diff
@@ -1,484 +1,167 @@
...
+    results_reconstruction["psnr"] = psnr_results
+    results_reconstruction["ssim"] = ssim_results
+    return results_reconstruction
+
+
+def compute_metrics(
+    blended_images,
+    isolated_images,
+    blend_list,
+    detection_catalogs=None,
```
ismael-mendoza, line 126: this should be a mandatory argument right?

btk/metrics.py
thuiop: Yeah exactly. We are matching each true object with the closest detected object, but this is not enough because we don't want to have 2 objects matched with the same detection, so we only keep the one which is the closest. This may not be the perfect way to do it since it can lead to cases were you have two detections and two objects but only one object matched...

btk/metrics.py
```diff
@@ -1,484 +1,279 @@
...
-            if len(q_true) > 0:
-                (q_det,) = np.where(summary_table[q_true, 1] == j)
-                eff_matrix[j, i] = len(q_det)
-    norm = np.sum(eff_matrix, axis=0)
-    # If no detections along a column, set sum to 1 to avoid dividing by zero.
-    norm[norm == 0.0] = 1
-    # normalize over columns.
-    eff_matrix = eff_matrix / norm[np.newaxis, :] * 100.0
-    return eff_matrix
+def detection_metrics(blended_images, isolated_images, blend_list, detection_catalogs, matches):
```
ismael-mendoza, line 61: I noticed some of these arguments are unused, is that intentional?

thuiop: Kind of, I just included all the possibly useful arguments in the functions ; I will probably remove them in the future.

btk/metrics.py
```diff
@@ -1,484 +1,279 @@
...
+            row["distance_closest"] = matches[i]["dist"][j]
+            if "reconstruction" in use_metrics:
+                row.add_columns([[0.0], [0.0], [0.0]], names=["mse", "psnr", "ssim"])
+                row["mse"] = results["reconstruction"]["mse"][i][j]
+                row["psnr"] = results["reconstruction"]["psnr"][i][j]
+                row["ssim"] = results["reconstruction"]["ssim"][i][j]
+            if "segmentation" in use_metrics:
+                row.add_column([0.0], name="iou")
+                row["iou"] = results["segmentation"]["iou"][i][j]
+            results["galaxy_summary"].add_row(row[0])
```
ismael-mendoza, line 249: Thanks @thuiop I liked what you did here and you are right, I think we can organize all the information into a single table that will contain all the information we need for any plots/summaries users might be interested in producing. 

Correct me if I'm wrong, but I think you combined all the blends in a batch into a single table right? Maybe we can add a column called `blend_id` to keep track of which galaxies where in which blend?

thuiop: I added the blend_id

btk/measure.py
```diff
@@ -218,18 +218,30 @@ def __next__(self):
...
+        for i, f in enumerate(self.measure_functions):
+            measure_dic = {}
+            for key in ["catalog", "deblended_images", "segmentation"]:
+                if measure_output[0][i][key] is not None:
+                    measure_dic[key] = [
+                        measure_output[j][i][key] for j in range(len(measure_output))
+                    ]
+            measure_results[f.__name__] = measure_dic
+        if len(self.measure_functions) == 1:
+            measure_results = measure_results[self.measure_functions[0].__name__]
```
ismael-mendoza, line 246: I see that you have two separate cases depending on how many measurement functions you are using (1 vs >1). This makes is so that we have to split the logic in the metrics code below (and possibly later in the unit tests). I would suggest getting rid of this additional 'if' statement for now. 

Later we can discuss how to make accessing the information less cumbersome for the user/us, but I think splitting the logic now just makes it confusing at least for me. (had to spend sometime figuring out `if` statement in `__next__` from metrics_generator)

btk/metrics.py
```diff
@@ -1,484 +1,351 @@
...
+        self, measure_generator, use_metrics=("detection"), meas_band_num=0, target_meas={}
+    ):
+        self.measure_generator = measure_generator
+        self.use_metrics = use_metrics
+        self.meas_band_num = meas_band_num
+        self.target_meas = target_meas
+
+    def __next__(self):
+        blend_results, measure_results = next(self.measure_generator)
+        if "catalog" not in measure_results.keys():
```
ismael-mendoza, line 335: It would be nice to avoid splitting the logic here (see comment above on `measure.py`.

btk/metrics.py
```diff
@@ -1,484 +1,351 @@
...
+class MetricsGenerator:
+    """Generator that calculates metrics on batches returned by the MeasureGenerator."""
+
+    def __init__(
+        self, measure_generator, use_metrics=("detection"), meas_band_num=0, target_meas={}
+    ):
+        self.measure_generator = measure_generator
+        self.use_metrics = use_metrics
+        self.meas_band_num = meas_band_num
+        self.target_meas = target_meas
```
ismael-mendoza, line 330: What is `target_meas` ? I actually couldn't figure it out from looking at its usage.

btk/metrics.py
```diff
@@ -1,484 +1,365 @@
...
+            precision.append(0)
+        recall.append(true_pos / (true_pos + false_neg))
+        if precision[-1] != 0 and recall[-1] != 0:
+            f1.append(2 / (1 / precision[-1] + 1 / recall[-1]))
+        else:
+            f1.append(0)
+    results_detection["precision"] = precision
+    results_detection["recall"] = recall
+    results_detection["f1"] = recall
+    return results_detection
```
ismael-mendoza, line 130: This values don't get added to the `galaxy_summary` table later because they are 'per blend' quantities right? 

Two things, (1) Would it be a good idea to have another table `blend_summary` with information for each blend? I can image this might simplify things for some users looking to summarize their results. (2) Is it useful to calculate these 3 quantities for each blend or should we just do it once for some large quantity of batches? Is the average `f1` over batches the same as calculating `f1` for all batches combined as a single dataset?

btk/measure.py
thuiop: I don't know, it doesn't feel too cumbersome to me ; but why not.

btk/metrics.py
thuiop: Basically, we want it to contain functions that measure a physical parameter on the images, which is then both ran on isolated_images and on deblended_images. Later, we can compare the two results and it gives an idea on how well the deblender performed relatively to the physical parameter you are interested in (currently only ellipticity but the structure is flexible)

btk/metrics.py
thuiop: Those are per blend quantities, but I thought some more and they are probably better to calculate on batch, and I don't believe you can just average them afterward ; I will get on it later.

thuiop:
Concerning the "run_metrics" function, I think we can include everything in the metrics generator directly ; or have a function that acts on the output of the compute_metrics function instead to make it compatible for users not using the generator.

btk/metrics.py
```diff
@@ -1,484 +1,407 @@
...
+    """Uses a `metrics_generator` objec to summarize metrics results for `n_batches` batches."""
+    measure_funcs = metrics_generator.measure_generator.measure_functions
+    summary_tables = {f: astropy.table.Table() for f in measure_funcs}
+    for i in range(n_batches):
+        blend_results, measure_results, metrics_results = next(metrics_generator)
+        for f in measure_funcs:
+            summary_tables[f] = astropy.table.vstack(
+                summary_tables[f], metrics_results["galaxy_summary"]
+            )
+    return summary_tables
```
ismael-mendoza, line 407: Here is a first draft of a function that uses info from multiple batches. I would also like to add overall results on all batches/blends. I guess it depends how we want to display the data...

ismael-mendoza: Definitely open to suggestions on this one, still thinking on how to summarize this in the best way

ismael-mendoza:
> Concerning the "run_metrics" function, I think we can include everything in the metrics generator directly ; or have a function that acts on the output of the compute_metrics function instead to make it compatible for users not using the generator.

ah just saw this after finishing writing the draft. Feel free to modify or add suggestions, I think for now it's fine if we operate on the generator but we can think about this more or discuss tomorrow.

btk/metrics.py
```diff
@@ -1,484 +1,497 @@
...
+    pixel_scale = additional_params["pixel_scale"]
+    meas_band_num = additional_params["meas_band_num"]
+    gal_image = galsim.Image(image[meas_band_num, :, :])
+    gal_image.scale = pixel_scale
+    try:
+        res = galsim.hsm.EstimateShear(gal_image, psf_image, shear_est=shear_est, strict=True)
+        result = [res.corrected_g1, res.corrected_g2, res.observed_shape.e]
+    except RuntimeError as e:
+        print(e)
+        result = [10.0, 10.0, 10.0]
```
ismael-mendoza, line 47: I wonder if we should use the same placeholder everywhere for consistency. Either `np.nan` or `-1` would be ok I think? Or did you have a reason to use `10.0` here?

btk/metrics.py
```diff
@@ -1,484 +1,497 @@
...
-        num_spurious1,
-        num_shred1,
-        num_detected2,
-        num_undetected2,
-        num_spurious2,
-        num_shred2,
-    ]
-    return blend_summary
 
+def detection_metrics(blend_list, detection_catalogs, matches):
```
ismael-mendoza, line 101: Thanks for making this function more "global" @thuiop - so now it's calculated using all the information from a single batch. I think we might want to make it even more "global" and calculate it only over several batches at once (so in `run_metrics`, again because we can't simply average all the f1-scores over all batches right?). 

I think we can do it by including a `detection_summary` table which for now we can just have columns `blend_id` and `matched`.

btk/metrics.py
ismael-mendoza: That makes sense, I added info in the docstring. Thanks!

btk/metrics.py
```diff
@@ -1,484 +1,497 @@
...
+    denom = np.sum(np.sum(blend_iso_images, axis=0) * iso_image)
+    return 1 - num / denom
 
-def get_closest_neighbor_distance(true_table):
-    """Returns a astropy.table.column with the distance to the closest object.
 
-    Function uses scipy.spatial to compute distance between the object centers.
-    If object is the only one in the blend then the `min_dist` value is set to
-    np.inf.
+def meas_ellipticity(image, additional_params, shear_est="KSB"):
```
aboucaud, line 27: as discussed during the telecon, maybe we should stick to KSB and rename the method `meas_ksb_ellipticity` for instance

btk/metrics.py
ismael-mendoza: I guess this depends on the `batch_size`, I imagine we probably want to keep the batch_size small, and use multiprocessing ?

ismael-mendoza: actually if we return the `true_pos`, `false_pos`, etc. for each blend/batch that should be enough?

btk/metrics.py
thuiop: No particular reason. I think the placeholder from galsim is actually -10.0 or something. -1 should be avoided as it is technically a possible value for the ellipticity.

btk/metrics.py
thuiop: If we return the raw numbers we can compute the metrics on several batches yes. Although I do not see why you would particularly want that compared to a single batch, except maybe if you want to compute the data in several times.

btk/metrics.py
```diff
@@ -293,8 +294,8 @@ def compute_metrics(  # noqa: C901
         segmentations (list) : Contains the measured segmentations, as a list of boolean arrays of
                                shape MHW where M is the number of detected objects (must be
                                consistent with corresponding detection catalog)
-        deblended_images (list) : Contains the deblended images, as a list of arrays of shape MCHW
-                                or MHWC depending on dim_order, where M is the number of detected
+        deblended_images (list) : Contains the deblended images, as a list of arrays of shape NCHW
```
thuiop, line 297: I'm not sure that I agree with that change. I put M instead of N there to differentiate with the number of blends which we usually call N (and which would be the length of the list here).

btk/metrics.py
```diff
@@ -1,484 +1,497 @@
...
-        results["shapes"].append(evaluate_shapes(shapes, index=i))
-    return results
+
+        self.blend_counter += len(blend_results["blend_list"])
+        return blend_results, measure_results, metrics_results
+
+
+def run_metrics(metrics_generator: MetricsGenerator, n_batches=100):
+    """Uses a `metrics_generator` objec to summarize metrics results for `n_batches` batches."""
+    measure_funcs = metrics_generator.measure_generator.measure_functions
```
aboucaud, line 489: that line feels weird..

aboucaud (commented):
Here are my general remarks on the current version of the metrics.

I have some more, but they belong to a future PR.

btk/measure.py
```diff
@@ -89,7 +89,7 @@ def sep_measure(batch, idx):
 
     image = batch["blend_images"][idx]
     stamp_size = image.shape[-2]  # true for both 'NCHW' or 'NHWC' formats.
-    coadd = np.mean(image, axis=0)
+    coadd = np.mean(image, axis=np.argmin(image.shape))  # Smallest dimension is the channels
```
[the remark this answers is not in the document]
thuiop: Yes, I tried to do it without having any additional information in input. Maybe adding the keyword would be better.

btk/measure.py
```diff
@@ -99,7 +99,11 @@ def sep_measure(batch, idx):
     for i in range(n_objects):
         seg_i = segmentation == i + 1
         segmentation_exp[i] = seg_i
-        deblended_images[i] = image * seg_i[np.newaxis, ...]
+        seg_i_reshaped = np.zeros((np.min(image.shape), stamp_size, stamp_size))
+        for j in range(np.min(image.shape)):
+            seg_i_reshaped[j] = seg_i
+        seg_i_reshaped = np.moveaxis(seg_i_reshaped, 0, np.argmin(image.shape))
```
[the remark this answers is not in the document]
thuiop: I think that one needs to be done to get the correct shape.

btk/plot_utils.py
```diff
@@ -379,3 +379,84 @@ def show_scarlet_residual(blend, observation, limits=(30, 90)):
...
+    for i, m in enumerate(metrics):
+        quantile = np.quantile(m, upper_quantile)
+        m_filtered = m[m <= quantile]
+        plt.hist(m_filtered, bins=bins, label=metric_names[i], alpha=0.6)
+        mean = np.mean(m_filtered)
+        plt.axvline(mean, linestyle="--", label="mean", color="blue")
+        median = np.median(m_filtered)
+        plt.axvline(median, linestyle="--", label="median", color="red")
+    plt.legend()
+    plt.show()
```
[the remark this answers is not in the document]
thuiop: That is probably a good idea, I was not sure how to approach it.

btk/metrics.py
```diff
@@ -1,484 +1,497 @@
...
-def run(metrics_params, test_size=1000):
-    """Runs detection/segmentation/flux/shape measurement algorithm defined in
-    the input metrics params for input test_size number of btk runs.
+    if dim_order == "NHWC":
+        blended_images = np.moveaxis(blended_images, -1, 1)
+        isolated_images = np.moveaxis(isolated_images, -1, 2)
+        if deblended_images is not None:
+            deblended_images = [np.moveaxis(im, -1, 1) for im in deblended_images]
+    elif dim_order != "NCHW":
+        raise ValueError("dim_order must be either 'NCHW' or 'NHWC'.")
```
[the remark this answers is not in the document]
thuiop: I am not sure if it is that useful to do it in a method. I see that you opened an issue regarding the dim_order, I suggest that we normalize it once this PR is closed.

btk/metrics.py
```diff
@@ -1,484 +1,497 @@
...
+                - "detection"
+                - "segmentation"
+                - "reconstruction"
+            meas_band_num (int): If using multiple bands for each blend,
+                which band index do you want to use for measurement?
+            target_meas (dict): Dictionary containing functions that can measure a physical
+                parameter on isolated galaxy images. Each key is the name of the estimator and
+                value the function performing the estimation (e.g. `meas_ellipticity` above).
+        """
+        self.measure_generator: MeasureGenerator = measure_generator
```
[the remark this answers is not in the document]
thuiop: We would have to rewrite the __next__  and __init__ method anyway, so I am not sure if it is actually useful.

btk/metrics.py
thuiop: Agreed. I don't know about this whole run_metrics function...

btk/metrics.py
ismael-mendoza: I think this is addressed by the `galaxy_summary`

btk/metrics.py
ismael-mendoza: I feel like the most common workflow is to have `batch_size` relatively small so that when doing multiprocessing each `cpu` doesn't have a lot of work to do in the `draw_blend_generator` and in the `measure_generator`. At the end multiple batches can be combined to get a large dataset.  Am I misunderstanding something? This is also why I implemented the `run_metrics` function.

Do let me know if you think an alternative workflow would be better though @thuiop

btk/metrics.py
ismael-mendoza: Yes, you are right, I hadn't read the description above. I just changed it back

btk/measure.py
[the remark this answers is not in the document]
ismael-mendoza: good point, I hand't realized this depended on "NCHW" or "NHWC" convention. Initially we wanted all `measure_functions` to only take in two arguments `batch` and `idx`, now I see that this might be too restrictive for some cases. 

One idea is to allow for `measure_functions` to have keywords and pass them in the initialization of the `measure_generator`. Another alternative is to use `partial`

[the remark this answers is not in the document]
ismael-mendoza: This can also be done in a separate PR addressing #133

btk/metrics.py
```diff
@@ -1,484 +1,497 @@
...
-    num_shred2 = len(np.where(true_table["num_detections2"] > 1)[0])
-    if not num_detected1 + num_undetected1 + num_shred1 == num_true:
-        raise ValueError(
-            "Number of detected objects + number undetected "
-            "objects must be equal to the total number of true "
-            "objects"
-        )
+    match_table = astropy.table.Table()
+    t_x = true_table["x_peak"].reshape(-1, 1) - detected_table["x_peak"].reshape(1, -1)
+    t_y = true_table["y_peak"].reshape(-1, 1) - detected_table["y_peak"].reshape(1, -1)
```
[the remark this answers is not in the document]
ismael-mendoza: Yes, there is still work todo in homogenizing the tables I think. Not sure of the best way to do this though. If the user uses the `measure_generator`, then that takes care of homogenization and does some checks

btk/metrics.py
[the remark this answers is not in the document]
ismael-mendoza: yeah this channel order issue is probably going to appear scattered in multiple cases in the code which is unfortunate... I echo @thuiop comment to leave this for a future PR

btk/metrics.py
thuiop: The batch size is not the size of the batches for multiprocessing ; when you set `cpus` to a value larger than 1, the batch is divided in several minibatches of size batch_size/cpus and then recombined as a single batch later. You may then want to generate several batches using the next method repeatedly if you want several batches.

ismael-mendoza: That's true. Anyways I think this is not too important, I'll let you handle how you think the best way for users to interact with the `metrics_generator` is. Feel free to remove the `run_metrics` function if you don't find it useful.

btk/measure.py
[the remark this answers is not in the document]
ismael-mendoza: I pushed a change to address this in PR #136 , feel free to review if you'd like @thuiop

btk/measure.py
```diff
@@ -99,7 +99,11 @@ def sep_measure(batch, idx):
     for i in range(n_objects):
         seg_i = segmentation == i + 1
         segmentation_exp[i] = seg_i
-        deblended_images[i] = image * seg_i[np.newaxis, ...]
+        seg_i_reshaped = np.zeros((np.min(image.shape), stamp_size, stamp_size))
```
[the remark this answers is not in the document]
ismael-mendoza: hopefully fixed in #136, let's discuss additional changes there.

btk/metrics.py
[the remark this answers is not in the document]
thuiop: After thinking about it, I am not sure if this is very interesting as the results from the `__next__` method are different in the MeasureGenerator and in the MetricsGenerator (the metrics generator returns the metrics as well), so this does not feel very natural.

thuiop:
I think we are getting toward the point where we could merge this PR, which is getting pretty big now. I will focus on plotting the results in another PR.

btk/draw_blends.py
```diff
@@ -243,9 +243,17 @@ def __next__(self):
...
+                    len(s.filters),
+                    pix_stamp_size,
+                    pix_stamp_size,
+                )
+                if not self.channels_last
+                else (
+                    pix_stamp_size,
+                    pix_stamp_size,
+                    len(s.filters),
+                )
```
ismael-mendoza, line 256: maybe avoid the command taking up several lines by declaring the variables separately? e.g.
```
shape1 = (len(s.filters), pix_sample_size, pix_stamp_size)
shape2 = pix_sample_size, pix_stamp_size, (len(s.filters))
image_shape = shape1 if not self.channels_last else shape2
```

btk/measure.py
```diff
@@ -1,35 +1,40 @@
...
-                                    order in the returned `catalog`. Where `n_objects` is the
-                                    number of detected objects
-    - segmentation (np.ndarray): Array of booleans with shape `(n_objects,stamp_size,stamp_size)`
-                                 The pixels set to True in the i-th channel correspond to the i-th
-                                 object. The order should correspond to the order in the returned
-                                 `catalog`.
+
+* catalog (astropy.table.Table): An astropy table containing measurement information. The
+  `len` of the table should be `n_objects`. If your
+  DrawBlendsGenerator uses a single survey, the following
```
ismael-mendoza, line 13: thanks for organizing the docstring so now it's clearer

btk/measure.py
ismael-mendoza: I will commit this change soon

btk/metrics.py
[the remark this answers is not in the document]
ismael-mendoza: I think it's fine to keep them separate. Conceptually they do different things...

ismael-mendoza (commented):
Looks really good @thuiop - almost ready to merge. Additional changes can be made later (diff will be more manageable then!) I just have to push the commit to simplify the logic. Thanks for your work on this !

btk/draw_blends.py
ismael-mendoza: did this in the latest commit

PR #129, opened 2021-04-15
Avoid conda in worflows

ismael-mendoza:


.github/workflows/test.yml
```diff
@@ -20,32 +17,37 @@ jobs:
...
-          conda install tensorflow==1.15
-          conda install -c conda-forge poetry
+          sudo apt-get install libfftw3-dev libeigen3-dev
 
       - name: Install poetry dependencies
-        shell: bash -l {0}
         run: |
+          poetry run pip install numpy Cython
+          poetry run pip install "eigency>=1.77"
           poetry install --extras "galsim-hub"
```
ismael-mendoza, line 48: I managed to get `poetry` up and running in the workflow again (no more conda!) . The only problem is `galsim` which fails if I don't install `numpy, Cython, eigency` before everything else. I'd be very interested in hearing a solution to this (but not urgent)

PR #131, opened 2021-04-16
Im/add requirements

ismael-mendoza:
- Correct problem with trailing-whitespace pre-commit and markdown files
- Adds `requirements.txt` for users who don't want to use poetry. 
- some small edits to `CONTRIBUTING.md`


ismael-mendoza:
Coverage went down because I just removed the `galsim_hub` unit-test. Will resolve with #135

Issue #127, opened 2021-04-07
versioning in BTK


ismael-mendoza:
- scm: automatic tools for versioning; you have to tag it.
- `pathlib` instead of hardcoded paths in unit-tests.

ismael-mendoza:
`pathlib` suggestion was implemented in #134

PR #132, opened 2021-04-21
Enable dynamic versioning of BTK

aboucaud:
Versioning is often difficult to handle as there are many steps necessary that can easily be skipped by inadvertance. 

Dynamic versioning is a way to automate the incrementation of project version corresponding to the developers needs.
It is based on a git tags rather that manual changes.

The plugin proposed here is called `poetry-dynamic-versioning` and can be found in
https://github.com/mtkennerly/poetry-dynamic-versioning

The PR also answers comments in #127

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -36,14 +36,12 @@ jobs:
           key: ${{ runner.os }}-${{ hashFiles('./poetry.lock') }}
 
       # galsim needs this.
-      - name: Instal FFTW + Eigen
+      - name: Instal FFTW
         run: |
-          sudo apt-get install libfftw3-dev libeigen3-dev
+          sudo apt-get install libfftw3-dev
 
       - name: Install poetry dependencies
         run: |
-          poetry run pip install numpy Cython
-          poetry run pip install "eigency>=1.77"
           poetry install --extras "galsim-hub"
 
       - name: build docs
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -26,6 +26,7 @@ jobs:
         run: |
           pip install --upgrade pip
           pip install poetry
+          pip install poetry-dynamic-versioning
 
       - name: Poetry cache
         uses: actions/cache@v2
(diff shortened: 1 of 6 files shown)


ismael-mendoza:
Thanks @aboucaud for this PR, dynamic versioning will definitely be useful for us. And I hadn't seen that plugging for poetry so thanks for finding it.

Could you add a small example or comments on how to use it in the `CONTRIBUTING.md` document if you don't mind? I know to tag in git all you have to do is something like
```
git tag -a v1.4 -m "my version 1.4"
```
does this plugin take care of the rest? I'm not too familiar with versioning myself so anything you can share in best practice either in this PR or in ``CONTRIBUTING.md` would be really useful :)

ismael-mendoza:
I looked more into it and I think looking at this resource is sufficient: https://semver.org along with the tagging code I added above. 

Would you recommend that every PR has a tag @aboucaud ?

ismael-mendoza:
I will fix the problem with the CI soon, it seem to be because there was an old version from before...

aboucaud:
So.. versoning.. complicated task !
I'll try to answer your comments.

First, I do not have many resources about it because it really depends on the mentality of the maintainer.
But now that you got a bit familiar with semver, the idea is 
```
MAJOR version when you make incompatible API changes,
MINOR version when you add functionality in a backwards compatible manner, and
PATCH version when you make backwards compatible bug fixes.
```
The tagging is only done on the `main` branch. So in general when you start using caring about versioning, you create a `develop` branch, on which the developers are sending the "feature PR's". As maintainer, you are the chief of orchestra, and select in which order PR's are merged on `develop`. 
When meaningful changes have been made, you create a PR from develop to main and then tag with an increasing minor version.   
When a milestone is reached, you can decide on a full major version bump/release of the code.  
But for quick fixes, or patches, you can merge any PR directly on the `main` and bump the patch version.

This should not go into `CONTRIBUTING.md` because it is only relevant for the person in charge of actually merging the files. All the developers need to care about is to send PR's to `develop` (which can be designated as the main target branch for ease of use).

Does that make sense ?

ismael-mendoza:
Thanks @aboucaud - I hope to adopt this framework soon then :)

ismael-mendoza:
Closes #127

ismael-mendoza:
Since it's been a while , I will now do this in a different branch

PR #134, opened 2021-04-22
Use pathlib to resolve path independently of OS

aboucaud:
Tentative of path resolution for Windows and Unix-like systems for the access of data files in tests

Answers #127

### DIFF
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -3,10 +3,10 @@ from conftest import data_dir
 from btk.catalog import CatsimCatalog
 from btk.catalog import CosmosCatalog
 
-CATALOG_PATH = (data_dir / "sample_input_catalog.fits").as_posix()
+CATALOG_PATH = str(data_dir / "sample_input_catalog.fits")
 COSMOS_CATALOG_PATHS = [
-    (data_dir / "cosmos/real_galaxy_catalog_23.5_example.fits").as_posix(),
-    (data_dir / "cosmos/real_galaxy_catalog_23.5_example_fits.fits").as_posix(),
+    str(data_dir / "cosmos/real_galaxy_catalog_23.5_example.fits"),
+    str(data_dir / "cosmos/real_galaxy_catalog_23.5_example_fits.fits"),
 ]
 
 
--- a/tests/test_catalog.py
+++ b/tests/test_catalog.py
@@ -1,10 +1,12 @@
+from conftest import data_dir
+
 from btk.catalog import CatsimCatalog
 from btk.catalog import CosmosCatalog
 
-CATALOG_PATH = "data/sample_input_catalog.fits"
+CATALOG_PATH = data_dir / "sample_input_catalog.fits"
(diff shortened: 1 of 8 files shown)


aboucaud:
I'd be glad if you could test this with poetry @ismael-mendoza

tests/test_error_cases.py
```diff
@@ -19,14 +21,12 @@ def compatible_catalogs(self):
...
     with pytest.raises(AttributeError) as excinfo:
-        catalog_name = "data/sample_input_catalog.fits"
-
         stamp_size = 24.0
         batch_size = 8
         cpus = 1
         add_noise = True
 
-        catalog = btk.catalog.CatsimCatalog.from_file(catalog_name)
+        catalog = btk.catalog.CatsimCatalog.from_file(CATALOG_PATH)
```
ismael-mendoza, line 29: It seems like the unit tests are failing in this function. Could you take a look @aboucaud ? You might also need to rebase because I just pushed a change to fix a separate issue in the unit-tests. (I would help out with this myself but since this is in a fork I can't make the changes)

ismael-mendoza (changes requested):
Thanks @aboucaud , this looks good to me. I think we can merge once tests are passing (and after you rebase your fork's branch with main)

tests/test_error_cases.py
aboucaud: You can because I gave you the rights as maintainer of the repo.

tests/test_catalog.py
```diff
@@ -1,10 +1,12 @@
...
 from btk.catalog import CatsimCatalog
 from btk.catalog import CosmosCatalog
 
-CATALOG_PATH = "data/sample_input_catalog.fits"
+CATALOG_PATH = (data_dir / "sample_input_catalog.fits").as_posix()
 COSMOS_CATALOG_PATHS = [
-    "data/cosmos/real_galaxy_catalog_23.5_example.fits",
-    "data/cosmos/real_galaxy_catalog_23.5_example_fits.fits",
+    (data_dir / "cosmos/real_galaxy_catalog_23.5_example.fits").as_posix(),
+    (data_dir / "cosmos/real_galaxy_catalog_23.5_example_fits.fits").as_posix(),
```
aboucaud, line 9: So about that change you made..the goal of using Path was to be able to handle both posix (Unix like) and Windows systems. By restricting it, it loses its interest. Could you put the Traceback of the error you had before changing it please ?

aboucaud: @ismael-mendoza could you try with calling `str()` on the Path rather than using `.as_posix()` ?

ismael-mendoza: oops you are right, I thought `as_posix()` was equivalent to `str()` but I guess not.

ismael-mendoza: I changed the calls to `str()` and the tests still pass

ismael-mendoza (approved):
Looks good, I think we can merge now

PR #136, opened 2021-04-22
[WIP] Use boolean to check input dimensions ordering

### LINKED ISSUE #133: `dim_order` possible simplification
In BTK at several places, one needs to know where the channels are last or not.
This is done through the `dim_order` keyword, that takes as input either "NCHW" or "NHWC".

I'm old enough to recall that issue existed in the early days of `tensorflow` because the default ordering had changed from the one used in `theano`, one of TF precursors.
They solved it by adding a boolean keyword called `channels_last`. 

IMHO we should use that convention as well, because
1. it spells easily and no one has to refer to the doc to understand its meaning
2. the boolean nature makes it very appropriate for the binary choice it does.

The only reason not to switch would be if you guys had in mind a third or more possibility of ordering the data.

aboucaud:
This PR replaces the use of `dim_order="NCHW` with `channels_last=False` and `dim_order="NHWC` with `channels_last=True`.

Fixes #133

### DIFF
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -147,7 +147,7 @@ class DrawBlendsGenerator(ABC):
         add_noise=True,
         shifts=None,
         indexes=None,
-        dim_order="NCHW",
+        channels_last=False,
     ):
         """Initializes the DrawBlendsGenerator class.
 
@@ -167,9 +167,9 @@ class DrawBlendsGenerator(ABC):
                            with indexes.
             indexes (list): Contains the ids of the galaxies to use in the stamp.
                         Must be of length batch_size. Must be used with shifts.
-            dim_order (str): Whether to return images as numpy arrays with the channel
-                                (band) dimension before the pixel dimensions 'NCHW' (default) or
-                                after 'NHWC'.
+            channels_last (bool): Whether to return images as numpy arrays with the channel
+                                (band) dimension as the last dimension or before the pixels
+                                dimensions (default).
         """
         self.blend_generator = BlendGenerator(
             catalog, sampling_function, batch_size, shifts, indexes, verbose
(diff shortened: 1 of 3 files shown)


ismael-mendoza:
Thanks @aboucaud for this PR. I pushed a change just now were I have a proposal of how to fix the issue you pointed out with the `measure_functions`. Please let me know what you think, I had to extend `btk.multiprocess` to accept `kwargs` which was the main change.

aboucaud:
Well I had in mind (and forgot to mention it in the meantime) was to wait until the PR #125 was merged to add more commits regarding the issue and the changes implied in `measure.py`. 
That could also leave some room to update multiprocessing in a different PR since it's unrelated, although necessary.
 
Just want to avoid things colliding in the middle. With those changes it will now make either PR difficult to merge first because of the concurrent changes.

ismael-mendoza:
I see, that makes sense. I think the conflict should be pretty small but I agree that there will be some - I can rollback the changes I added and put them in separate PR  (to be merged when #125 is merged)

aboucaud:
I think it would be best, even for the sake of a clean history

aboucaud:
Also your work on multiprocessing is more important than this small refactoring PR

ismael-mendoza (approved):
Ok, I reverted the changes and this looks good to me now @aboucaud so I'm merging

aboucaud:
There was a misunderstanding. I might not have been clear enough in my explanation, sorry. 

I changed the PR name to _work in progress_ : [WIP] so that we could wait for PR #125 to be merged to `main`, then rebase and continue what you wanted to do. That way, there are no conflicts to be solved on either branch ; whereas now the #125 cannot be merged as it is.

ismael-mendoza:
ah I see what you mean now sorry about the confusion - hopefully it should still be quite smooth to the rebasing since the two PRs I merged did not touch `metrics.py` (and this one only touched measure.py in a comment)

PR #140, opened 2021-04-28
Fix measure channel last

ismael-mendoza:
This fixes a comment from @aboucaud in the most recent PR #125 and allows measurement functions to take an arbitrary number of keyword arguments. One downside is that the skeleton changed slightly, it is now required to be: 

 ```
def measure_function(batch, idx, **kwargs):
    # do some measurements on the images contained in batch.
    return output
 ```

since all the measure_functions share the keyword arguments passed in to the measure_generator initializer, so some measure_functions will get passed in keyword arguments that are not used. (open to suggestions)

### DIFF
--- a/btk/measure.py
+++ b/btk/measure.py
@@ -181,11 +181,12 @@ class MeasureGenerator:
         self.cpus = cpus
 
         self.batch_size = self.draw_blend_generator.batch_size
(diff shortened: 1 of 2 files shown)


thuiop (approved):

ismael-mendoza:
Thanks @thuiop for the review

PR #147, opened 2021-05-03
Added possibility to change how distance is computed

### LINKED ISSUE #144: Adding a way to change how the distance is calculated in metrics
Currently, we make matches only based on the center to center distance between galaxies, which does not take into account the size of the galaxies. It would be interesting to reimplement the possibility to scale the distance with the size of the galaxy, or even allow more complicated things.

thuiop:
This feature was missing from the metrics PR. Closes #144 

It will probably need to be showcased in a tutorial later but I will probably do that later.

### DIFF
--- a/btk/metrics.py
+++ b/btk/metrics.py
@@ -579,7 +579,6 @@ def compute_metrics(  # noqa: C901
                          - galaxy_summary : astropy Table containing all the galaxies from all
                            blends and related metrics
     """
-    print(f_distance)
     if channels_last:
         blended_images = np.moveaxis(blended_images, -1, 1)
         isolated_images = np.moveaxis(isolated_images, -1, 2)
--- a/btk/metrics.py
+++ b/btk/metrics.py
@@ -105,7 +105,21 @@ def meas_ksb_ellipticity(image, additional_params):
     return result
 
 
-def get_detection_match(true_table, detected_table):
+def distance_center(true_gal, detected_gal):
+    """Computes distance between the two galaxies given as arguments.
+
+    Args:
+        true_gal (astropy.table.Table): Contains information related to the true galaxy.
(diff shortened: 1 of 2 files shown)


btk/metrics.py
```diff
@@ -104,7 +104,21 @@ def meas_ksb_ellipticity(image, additional_params):
...
+        detected_gal (astropy.table.Table): Contains information related to the detected galaxy
+    Returns:
+        Distance between the two galaxies
+    """
+    return np.hypot(
+        true_gal["x_peak"] - detected_gal["x_peak"], true_gal["y_peak"] - detected_gal["y_peak"]
+    )
+
+
+def get_detection_match(true_table, detected_table, f_distance=None):
```
ismael-mendoza, line 121: why not use `f_distance=distance_center` in the keyword argument here and avoid the `if/else` statement below?

ismael-mendoza (dismissed):
Looks really good, just one small question

btk/metrics.py
ismael-mendoza: I guess the original method of doing it will be faster, although the for loop should always be small even with blends of ~8 unless the detection algorithm is really bad. Not sure if it's worth keeping the original method for sake of speed

thuiop: That's a good point, I will change it.

btk/metrics.py
```diff
@@ -553,14 +578,16 @@ def compute_metrics(  # noqa: C901
                          - galaxy_summary : astropy Table containing all the galaxies from all
                            blends and related metrics
     """
+    print(f_distance)
```
ismael-mendoza, line 581: I think you meant to remove this print statement

ismael-mendoza (changes requested):
I think you just missed a print statement, see below

ismael-mendoza (approved):

PR #148, opened 2021-05-03
Added possibility to save results

### LINKED ISSUE #142: Add a functionnality to export the measurement results
Once the metrics have been computed on the blends, it would be nice to have a way to export the results to a table/file so they can be loaded in `pandas` and results could be compared between runs.

Something like a `.to_csv()` or `to_dataframe()`.

thuiop:
Closes #142 

I still have to add utility functions for loading the saved results easily.

### DIFF
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -173,7 +173,7 @@ class DrawBlendsGenerator(ABC):
                                 (band) dimension as the last dimension or before the pixels
                                 dimensions (default).
             save_path (str): Path to save the results, ending by the file name root. If left
-                            ton None, results will not be saved.
+                            as None, results will not be saved.
         """
         self.blend_generator = BlendGenerator(
             catalog, sampling_function, batch_size, shifts, indexes, verbose
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -1,5 +1,6 @@
 """Module for generating batches of drawn blended images."""
 import copy
+import os
 from abc import ABC
 from abc import abstractmethod
 from itertools import chain
@@ -279,11 +280,16 @@ class DrawBlendsGenerator(ABC):
                 batch_blend_cat[s.name].append(batch_results[i][2])
 
             if self.save_path is not None:
-                np.save(f"{self.save_path}_{s.name}_blended", blend_images[s.name])
-                np.save(f"{self.save_path}_{s.name}_isolated", isolated_images[s.name])
+                if not os.path.exists(os.path.join(self.save_path, s.name)):
+                    os.mkdir(os.path.join(self.save_path, s.name))
+
+                np.save(os.path.join(self.save_path, s.name, "blended"), blend_images[s.name])
+                np.save(os.path.join(self.save_path, s.name, "isolated"), isolated_images[s.name])
                 for i in range(len(batch_results)):
                     batch_blend_cat[s.name][i].write(
-                        f"{self.save_path}_{s.name}_blend_info_{i}", format="ascii", overwrite=True
+                        os.path.join(self.save_path, s.name, f"blend_info_{i}"),
+                        format="ascii",
+                        overwrite=True,
                     )
 
         if len(self.surveys) > 1:
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -148,6 +148,7 @@ class DrawBlendsGenerator(ABC):
         shifts=None,
         indexes=None,
         channels_last=False,
+        save_path=None,
     ):
         """Initializes the DrawBlendsGenerator class.
 
@@ -170,6 +171,8 @@ class DrawBlendsGenerator(ABC):
             channels_last (bool): Whether to return images as numpy arrays with the channel
                                 (band) dimension as the last dimension or before the pixels
                                 dimensions (default).
+            save_path (str): Path to save the results, ending by the file name root. If left
+                            ton None, results will not be saved.
         """
         self.blend_generator = BlendGenerator(
             catalog, sampling_function, batch_size, shifts, indexes, verbose
@@ -191,8 +194,8 @@ class DrawBlendsGenerator(ABC):
 
         self.add_noise = add_noise
         self.verbose = verbose
-
         self.channels_last = channels_last
+        self.save_path = save_path
 
     def __iter__(self):
         """Returns iterable which is the object itself."""
@@ -274,6 +277,15 @@ class DrawBlendsGenerator(ABC):
                 blend_images[s.name][i] = batch_results[i][0]
                 isolated_images[s.name][i] = batch_results[i][1]
                 batch_blend_cat[s.name].append(batch_results[i][2])
+
+            if self.save_path is not None:
+                np.save(f"{self.save_path}_{s.name}_blended", blend_images[s.name])
+                np.save(f"{self.save_path}_{s.name}_isolated", isolated_images[s.name])
+                for i in range(len(batch_results)):
+                    batch_blend_cat[s.name][i].write(
+                        f"{self.save_path}_{s.name}_blend_info_{i}", format="ascii", overwrite=True
+                    )
+
         if len(self.surveys) > 1:
             output = {
                 "blend_images": blend_images,
@@ -291,6 +303,7 @@ class DrawBlendsGenerator(ABC):
                 "psf": psfs[survey_name],
                 "wcs": wcss[survey_name],
             }
+
         return output
 
     def render_mini_batch(self, blend_list, psf, wcs, survey, extra_data=None):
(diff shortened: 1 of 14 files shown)


thuiop:
This PR is ready to be reviewed ; the only shortcoming here is that PSF and WCS are not saved with the blends as galsim and astropy do not provide utility functions to do it cleanly. Should we use something like pickle to save it anyway ?

ismael-mendoza (changes requested):
Looks very good @thuiop, thanks for adding this functionality. Please see below for a couple of small requested changes and questions. It appears to me that it's really close to being ready for merge

btk/measure.py
```diff
@@ -268,11 +272,21 @@ def __next__(self):
...
                         measure_output[j][i][key] for j in range(len(measure_output))
                     ]
             measure_results[f.__name__] = measure_dic
+            if self.save_path is not None:
+                for key in ["segmentation", "deblended_images"]:
+                    if key in measure_dic.keys():
+                        np.save(f"{self.save_path}_{f.__name__}_{key}", measure_dic[key])
+                for j, cat in enumerate(measure_dic["catalog"]):
+                    cat.write(
+                        f"{self.save_path}_{f.__name__}_detection_catalog_{j}",
```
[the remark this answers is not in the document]
aboucaud: +1

tests/test_save.py
```diff
@@ -0,0 +1,57 @@
...
+
+import numpy as np
+
+import btk
+from btk.survey import Rubin
+
+
+def test_save():
+    try:
+        shutil.rmtree("./output")
```
aboucaud, line 12: I suggest not to use "output" as the test directory since it would be a default choice for a given user to name an actual output repo. Maybe something like `"output_test_dir".

Also I would like to point out a useful library for these kind of things : [tempfile](https://docs.python.org/3/library/tempfile.html)

btk/utils.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Returns:
+        Dictionnary containing the blend images, the isolated images and the
+        informations about the blends.
+    """
+    blend_images = np.load(f"{path}_{survey}_blended.npy", allow_pickle=True)
+    isolated_images = np.load(f"{path}_{survey}_isolated.npy", allow_pickle=True)
+    blend_list = []
+    for i in range(blend_images.shape[0]):
+        blend_list.append(Table.read(f"{path}_{survey}_blend_info_{i}", format="ascii"))
```
aboucaud, line 23:
```suggestion
    blend_list = [Table.read(f"{path}_{survey}_blend_info_{i}", format="ascii")
                          for i in range(blend_images.shape[0]]
```

btk/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        except FileNotFoundError:
+            print(f"No {key} found.")
+    catalog = []
+    for j in range(n_batch):
+        catalog.append(
+            Table.read(
+                f"{path}_{measure_name}_detection_catalog_{j}",
+                format="ascii",
+            )
+        )
```
aboucaud, line 59:
```suggestion
    catalog = [Table.read(f"{path}_{measure_name}_detection_catalog_{j}", format="ascii")
                      for j in range(n_batch)]
```

btk/utils.py
```diff
@@ -0,0 +1,124 @@
...
+        The three dictionnaries corresponding to the results.
+    """
+    blend_results = {}
+    for key in ["blend_images", "isolated_images", "blend_list"]:
+        blend_results[key] = {}
+    measure_results = {}
+    metrics_results = {}
+    for s in surveys:
+        blend_results_temp = load_blend_results(path, s)
+        for key in ["blend_images", "isolated_images", "blend_list"]:
```
aboucaud, line 117: redundant with line 111

aboucaud (commented):
I agree with @ismael-mendoza, one should be more attentive to how the `save_path` is constructed.

The way you wrote it, the user does not really have a choice for the filenames (probably a good thing), but it must have a choice of the main output directory then. If this is what you mean by `save_path` then we should use `pathlib` to construct the full paths. 
Also it may be easier (we can discuss this) to create subdirectories inside the `save_path` to put the various outputs, e.g. `save_path/survey/blended.npy` rather than `<save_path>_<survey>_blended.npy`. That way it is easier to check the directory existence rather than the existence of each file before loading.

btk/measure.py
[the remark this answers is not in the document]
thuiop: It should be a path ending by a filename root (eg `/path/to/dir/filename_{f.__name__}_detection_catalog_{j}`). And yes pathlib is probably better, but I was not sure how to do it with the filename root. I guess we can settle with just a folder.

btk/utils.py
thuiop: Ok, but I'm curious to what difference it makes.

aboucaud: Just a common refactoring practice.  
For one it specifies this is constant (and therefore a tuple would be even better than the list I proposed).  
And also, if in the future you want to add or remove an element, you have to do it once and limit the possibilities for typos.

thuiop:
I should have addressed all the comments.

btk/draw_blends.py
```diff
@@ -274,6 +278,20 @@ def __next__(self):
                 blend_images[s.name][i] = batch_results[i][0]
                 isolated_images[s.name][i] = batch_results[i][1]
                 batch_blend_cat[s.name].append(batch_results[i][2])
+
+            if self.save_path is not None:
+                if not os.path.exists(os.path.join(self.save_path, s.name)):
+                    os.mkdir(os.path.join(self.save_path, s.name))
```
ismael-mendoza, line 284: This makes sense to me, so now `save_path` is a directory which will contain one folder for each survey containing the images for that survey.

btk/draw_blends.py
```diff
@@ -170,6 +172,8 @@ def __init__(
             channels_last (bool): Whether to return images as numpy arrays with the channel
                                 (band) dimension as the last dimension or before the pixels
                                 dimensions (default).
+            save_path (str): Path to save the results, ending by the file name root. If left
+                            as None, results will not be saved.
```
ismael-mendoza, line 176: I think now this description is outdated? See below for suggestion

```suggestion
            save_path (str): Path to a directory where results will be saved. If left
                            as None, results will not be saved.
```

btk/measure.py
```diff
@@ -163,6 +165,8 @@ def __init__(
             verbose (bool): Whether to print information about measurement.
             measure_kwargs (dict): Dictionary containing keyword arguments to be passed
                 in to each of the `measure_functions`.
+            save_path (str): Path to save the results, ending by the file name root. If left
+                            ton None, results will not be saved.
```
ismael-mendoza, line 169: Same as above

```suggestion
            save_path (str): Path to a directory where results will be saved. If left 
                              as None, results will not be saved.
```

btk/measure.py
```diff
@@ -268,11 +273,24 @@ def __next__(self):
...
                         measure_output[j][i][key] for j in range(len(measure_output))
                     ]
             measure_results[f.__name__] = measure_dic
+            if self.save_path is not None:
+                if not os.path.exists(os.path.join(self.save_path, f.__name__)):
+                    os.mkdir(os.path.join(self.save_path, f.__name__))
+
+                for key in ["segmentation", "deblended_images"]:
+                    if key in measure_dic.keys():
+                        np.save(os.path.join(self.save_path, f.__name__, key), measure_dic[key])
```
ismael-mendoza, line 288: Thanks for introducing the `os.path.join` which I believe should be equivalent to using `pathlib`.

btk/metrics.py
```diff
@@ -542,6 +545,8 @@ def compute_metrics(  # noqa: C901
                              be returned for both isolated and deblended images to compare.
         channels_last (bool) : Indicates whether the images should be channels first (NCHW)
                           or channels last (NHWC).
+        save_path (str): Path to save the results, ending by the file name root. If left
+                            ton None, results will not be saved.
```
ismael-mendoza, line 549: Same suggestion as above

```suggestion
        save_path (str): Path to directory where results will be saved. If left
                      as None, results will not be saved.
```

btk/metrics.py
```diff
@@ -658,12 +671,15 @@ def __init__(
                 applied when getting segmentations from true images. A value of 3 would
                 correspond to a threshold of 3 sigmas (with sigma the standard deviation of
                 the noise)
+            save_path (str): Path to save the results, ending by the file name root. If left
+                            ton None, results will not be saved.
```
ismael-mendoza, line 675: Same as above
```suggestion
            save_path (str): Path to directory where results will be saved. If left 
                    as None, results will not be saved.
```

ismael-mendoza (approved):
I think this is ready to merge after the CI passes and if my commit suggestions below are indeed what you intended.

btk/measure.py
aboucaud: indeed, it's the old way, but for this case it works well

ismael-mendoza:
Thanks for resolving the merge conflicts @thuiop. Merging now -

PR #149, opened 2021-05-14
Added average precision (AUC) metric

### LINKED ISSUE #146: Add Area under Curve metric
An additionnal metric which was not implemented in the metrics PR is the area under curve metric. The idea is that an algorithm using a certain threshold as a parameter can be evaluated by plotting the precision-recall curve (sometimes called the ROC curve) for different values of the threshold and computing the area under the curve. As there is a tradeoff between precision and recall, an algorithm with a larger AuC (Area under Curve) will be overall better than algorithm with a lower one, as in average it will have a better precision and/or recall.
We already have prediction and recall computed in the metrics part, but getting the curve means that the algorithm should be ran several times with different thresholds, which is more something from the measure part. 
My take on how to implement this would be to have the possibility for users to pass a list of kwargs for the measure function in the measure generator, and the function would be executed for each element of the list. We can then recover the precision and recall for each different measure function in the metrics part, and have a separate function (outside compute_metrics) which could be used to get the curve and compute the AuC.
Another possibility is to expect users to provide the different measure functions with different parameters and only provide the function to compute the AuC and some guidelines on how to do it.
What do you think @ismael-mendoza ?

thuiop:
Closes #146.

I implemented the approach described in the issue. This changes the following things : 
- measure_kwargs should be provided as a list of dictionaries instead of a dictionary (the measure function will be executed one time with each dictionary in the list)
- the end results are named `sep_measure0, sep_measure1...`, even in the case where only one or no kwargs were provided.

When the user provides multiple kwargs, he can give the metrics results to the `auc` function to compute the average precision and optionally plot the precision-recall curve.

### DIFF
--- a/btk/measure.py
+++ b/btk/measure.py
@@ -274,12 +274,12 @@ class MeasureGenerator:
             if self.verbose:
                 print(f"Measurement {m} performed on batch")
             for i, f in enumerate(self.measure_functions):
-                measure_dic = {}
+                measure_dict = {}
                 for key in ["catalog", "deblended_images", "segmentation"]:
                     if measure_output[0][i][key] is not None:
-                        measure_dic[key] = [
+                        measure_dict[key] = [
                             measure_output[j][i][key] for j in range(len(measure_output))
                         ]
-                measure_results[f.__name__ + str(m)] = measure_dic
+                measure_results[f.__name__ + str(m)] = measure_dict
 
         return blend_output, measure_results
--- a/btk/measure.py
+++ b/btk/measure.py
@@ -103,6 +103,7 @@ def sep_measure(batch, idx, channels_last=False, **kwargs):
     """
     if isinstance(batch["blend_images"], dict):
         raise NotImplementedError("This function does not support the multi-resolution feature.")
+    sigma_noise = kwargs.get("sigma_noise", 1.5)
 
     image = batch["blend_images"][idx]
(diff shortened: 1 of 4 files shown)


btk/measure.py
```diff
@@ -254,25 +260,26 @@ def __next__(self):
...
+                self.run_batch,
+                args_iter,
+                kwargs_iter=kwargs_iter,
+                cpus=self.cpus,
+                verbose=self.verbose,
+            )
+            if self.verbose:
+                print(f"Measurement {m} performed on batch")
+            for i, f in enumerate(self.measure_functions):
+                measure_dic = {}
```
ismael-mendoza, line 277: change `measure_dic` -> `measure_dict`

btk/metrics.py
```diff
@@ -699,3 +700,57 @@ def __next__(self):
             metrics_results[meas_func] = metrics_results_f
 
         return blend_results, measure_results, metrics_results
+
+
+def auc(metrics_results, measure_name, n_meas, plot=False, ax=None):
```
ismael-mendoza, line 705: This looks great 👍

ismael-mendoza (dismissed):
Looks good to me, couldn't really find anything wrong :) Thanks for adding this useful metric

ismael-mendoza (approved):

PR #150, opened 2021-05-19
Rework CLI

### LINKED ISSUE #48: btk_input is broken and cannot be currently used
Comment from Alexandre: 

> the way to programmatically run BTK without running into too much details is through the btk_input.py which is currently outdated. This script should be incorporated into the module and be made available to the user upon install via as a command-line script through and entry_point in the setup.py

ismael-mendoza:
Closes #48 

I implemented a basic version of the CLI, we are missing features requested in previous BTK meeting and specified in #170 which will be implemented in future PRs (it was getting quite big and hard to review). 

- Run BTK from the command line after pip installing like 

```
btk batch_size=24 cpus=3 save_path=path/to/save/dir ...
```

for more details take a look at `conf/hydra/help/btk_help.yaml`.

- CLI runs everything from end-to-end, creates catalog, sampling_function, surveys, draw_blend_generator, measure_generator, and finally a metrics generator. We take advantage of the `save_path` from #148 to save everything automatically when running `next(metrics_generator)`. 

## Notes
- Main function is inside`main.py` but CLI is technically implemented in `__main__.py` (useful for unit testing)
- I decided to keep hydra for now (instead of switching completely to the galsim parser) because hydra has some nice features that saved me some work. I'm open to suggestions and changing away from hydra in the future though.
- Users who want to use the BTK CLI will need to create an environment variable `BTK_HOME` which is the absolute path to their local btk repo (has to do with the way hydra works).


ismael-mendoza:
This is ready for review, reviewers: I think the main thing to look at is the `btk/main.py` file (actually runs the CLI) and the `btk/hydra/help/btk_help.yaml` file (this file explains how to command line is supposed to be used). 

Please let me know if you have any questions and I'll try my best to clarify!

btk/main.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+
+    # get measure_functions.
+    measure_functions = []
+    f_names = cfg.measure.measure_functions
+    if not isinstance(f_names, Iterable):
+        f_names = [f_names]
+    for f_name in f_names:
+        if f_name not in available_measure_functions:
+            raise ValueError("Measure functions specified are not implemented in BTK.")
```
thuiop, line 53: For measure functions, we will definitely need for them to be customizable.

conf/hydra/help/btk_help.yaml
```diff
@@ -0,0 +1,96 @@
...
+  The CLI is designed to run BTK from start to end, starting by creating a
+  DrawBlendGenerator, followed by a MeasureGenerator, and finally calculating
+  and saving the corresponding metrics.
+
+  Assuming that BTK has been pip installed, you can run btk from the command line like e.g.
+
+  btk sampling=default draw_blends=catsim max_number=3 save_path=/directory/to/save/results cpus=1
+    verbose=False surveys=[Rubin, HST] sampling=random catalog.name=catsim
+    use_metrics=['detection', 'segmentation'] (...)
+
```
thuiop, line 34: We should also be able to write a config file and run it from there, right ? I may be useful to add a line on that.

thuiop (approved):
This PR seems pretty good overall, I have not spotted any important error.

btk/main.py
ismael-mendoza: Yes! this will be the next PR and I have it also written down in #170

conf/hydra/help/btk_help.yaml
ismael-mendoza: Yes, kinda of - it's a little difficult for the user to provide their own config from scratch actually. I was hoping to add all the required flexibility and useful defaults (in upcoming PRs) to the CLI directly so that creating a config file from scratch is unnecessary for the user.  

We could also provide a bash file that has all the defaults and that users can easily modify to their needs, but maybe let's discuss this more in the upcoming meeting and merge this for now if that's ok with you

Issue #89, opened 2021-01-31
Stack notebooks


ismael-mendoza:
Following #87, Scarlet and LSST Stack will no longer be a dependence. Instead, we will create notebooks/documentation that specifically talk about how to use these tools in BTK.

ismael-mendoza:
SCARLET notebook implemented in #151

ismael-mendoza:
deprecated

PR #151, opened 2021-05-19
Added SCARLET notebook

thuiop:
As mentioned in #89, the SCARLET implementation in BTK was broken and we don't want to have the scarlet package as a dependency. I wrote a notebook reusing some code from the old implementation to create a measure function using scarlet. Please note that scarlet should not be installed via pip (that version is largely outdated) and instead cloned and installed directly from GitHub.
The notebook is functional though it is yet devoid of details ; I will add some text later to explain what is going on in the tutorial.
This should also help for testing the plotting features with something more robust than the sep measure function.

I would like to get a review from someone who has already worked with scarlet, maybe @herjy ? It should be pretty straightforward as there is not a lot of code attached to it.


notebooks/scarlet-measure.ipynb
```diff
@@ -0,0 +1,2888 @@
+{
```
ismael-mendoza, line 1: <p>Maybe we should filter out this error since it can get really annoying for users doing a large sequence of measurements? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LSSTDESC/BlendingToolKit/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='6'/>

ismael-mendoza, line 1: <p>For the last plot, you might want to increase <code>figsize</code>so that values are readable? </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LSSTDESC/BlendingToolKit/pull/151/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='7'/>

thuiop: <p>I think this is the error from the ellipticity measurement in the metrics.</p>

thuiop: <p>Probably, this is bound to change with the plotting PR anyway.</p>

ismael-mendoza: OK let's ignore for now and revisit later then, at least we have this notebook already which is great

herjy (changes requested):
I reviewed the notebook. A little text to hold the user's hand could be helpful. 
The biggest issue I spotted is how the weights are set. 
A few more things could be improved around the channels setup and getting the models.

thuiop:
I will add a bit more text. Do you have more precise comments about the weights and the improvements you thought about ? If you made comments, I cannot see them.

herjy:
I've put the detailed comments on 
https://app.reviewnb.com/LSSTDESC/BlendingToolKit/blob/new-scarlet-notebook/notebooks%2Fscarlet-measure.ipynb%2F/discussion

herjy:
Tell me if you can't see them

thuiop:
Ok thank you, I could not find them for some reason. I will reply on reviewNB.

herjy:
This was a rather lose comments intend as "In case you need it". No need to act upon this.
Hmmm I haven't seen that error before. Submitting an issue in scarlet is always an option. It looks like it's just a warning though? A common culprit is when no source is detected.

herjy:
Appologies, I was looking at the wrong "error" [aka not the error at all]. Hmm it looks like changing to sources was not the right thing to do. I might have misread what you are doing here. For the time being, keep iterating over the `components` in `blend`.

thuiop:
The sources did seem like the right thing ; I think I already had the error when iterating with the blend. This might be happening only in specific situations, eg when galaxies are poorly resolved against the background (just an hypothesis). Maybe I can add a catch for the error to prevent it from stopping large batches ?

Anyway, we might have to refine it later ; for instance to support multiresolution ; which SCARLET should be able to doL.

herjy:
I did not get the error when running with blend, but I see, this might be coincidental. Something that regularly messes upscarlet is empty blends, aka images where no galaxy is detected. It might be something like that here.
Multiresolution should work in the exact same way as single resolution from the point of view of blends, sources and components.

thuiop:
It should be fine now ; I will add multiresolution when I get the metrics part working with the multiresolution.

ismael-mendoza (approved):

PR #155, opened 2021-05-20
Fix Intro notebook

ismael-mendoza:
Let me know if you have other suggestions or things that I missed, thanks! You can take a look at the notebook directly in the `ReviewNB` link below

I also fixed the issue with restricting BTK to python 3.7; instead any python can be used but the optional dependencies of galsim_hub are specified to be only for python 3.7 (see `pyproject.toml` diff)


thuiop (approved):
Seems good to me, as long as it runs. The batch size is too small to see anything from the metrics part though.

PR #159, opened 2021-05-21
Normalize inputs and outputs

### LINKED ISSUE #152: Do we need both `meas_bands` in DrawBlendGenerator and `meas_band_num` in MetricsGenerator?
They seem redundant, also I think we need the one in the metrics to be a tuple as well ? (in the case of multiple surveys?)

### LINKED ISSUE #153: Normalize input and output
Now that the structure of BTK is mostly fixed, we should address some issues regarding the input and the outputs.
- The output for multiple surveys is like `blend_results["blend_images"]["LSST"]` while the output for several measure functions is like `measure_results["sep_measure0"]["deblended_images"]`. I think we should swap the order for the surveys.
- We should think about how to handle the case where only one survey/measure function/measure_kwargs is inputted. Right now, surveys must be given as a list, but does not take the "dictionary form" when there is only one survey. For measure functions, the input is always a list and the output is always a dictionary index by the names of the functions. For the measure_kwargs, the input is a list (or None) and the output adds numbers to the end of the measure function name (hence the "sep_measure0"), even when the input is None or a list of length one. I personally feel that the best would be to have the possibility to give a single element and get the results in the "simple form", or give a list of elements and get the results in the "dictionary form" ; for the measure_kwargs this would mean not adding the zero in the name ; it would add some code though to do all the checks though.
- As pointed in #152 , the meas_bands are not really handled correctly right now
- There may be other redundant/useless parameters

thuiop:
Closes #153 and closes #152.

### DIFF
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -209,14 +209,9 @@ class DrawBlendsGenerator(ABC):
             output : Dictionary with blend images, isolated object images, blend catalog,
             PSF images and WCS.
         """
-        batch_blend_cat, batch_obs_cond = {}, {}
-        blend_images = {}
-        isolated_images = {}
         blend_cat = next(self.blend_generator)
         mini_batch_size = np.max([self.batch_size // self.cpus, 1])
-        psfs = {}
-        wcss = {}
-
+        output_dict = {}
         for s in self.surveys:
             pix_stamp_size = int(self.stamp_size / s.pixel_scale)
 
@@ -240,12 +235,6 @@ class DrawBlendsGenerator(ABC):
                         f"function nor a galsim object"
                     )
             wcs = make_wcs(s.pixel_scale, (pix_stamp_size, pix_stamp_size))
-            psfs[s.name] = psf
-            wcss[s.name] = wcs
-
-            # allocate memory for output catalogues and images.
-            batch_blend_cat[s.name] = []
-            batch_obs_cond[s.name] = []
 
             # decide image_shape based on channels_last bool.
             option1 = (len(s.filters), pix_stamp_size, pix_stamp_size)
@@ -253,8 +242,6 @@ class DrawBlendsGenerator(ABC):
             image_shape = option1 if not self.channels_last else option2
 
             # create emtpy arrays with image_shape.
-            blend_images[s.name] = np.zeros((self.batch_size, *image_shape))
-            isolated_images[s.name] = np.zeros((self.batch_size, self.max_number, *image_shape))
 
             input_args = []
             for i in range(0, self.batch_size, mini_batch_size):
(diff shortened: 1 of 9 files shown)


thuiop:
I ended up reversing the order for the measure functions, so that you access the results with `results["catalog"]["sep_measure"]` just like with the surveys (I realized that having the surveys in that order will help a lot for expanding the multiresolution feature to the metrics part). Also, you can now provide a survey not as a list, and get the "simple" form, and the same for the measure_function except that you always get the dictionary form ; this is not ideal, but it would add a lot of code in the metrics part. However, it does not add a 0 anymore when there is only one (or zero) measure_kwargs.

btk/draw_blends.py
```diff
@@ -41,28 +42,6 @@ def get_center_in_pixels(blend_catalog, wcs):
     return dx_col, dy_col
 
 
-def get_size(catalog, psf, pixel_scale):
```
ismael-mendoza, line 44: Did you remove this because you think it was unnecessary later on? I agree if that's the case

btk/measure.py
```diff
@@ -72,7 +74,7 @@ def basic_measure(batch, idx, channels_last=False, **kwargs):
     Returns:
             dict containing catalog with entries corresponding to measured peaks.
     """
-    if isinstance(batch["blend_images"], dict):
+    if "blend_images" not in batch.keys():
```
ismael-mendoza, line 77: Doesn't this always evaluate to false?

btk/measure.py
```diff
@@ -102,7 +104,7 @@ def sep_measure(batch, idx, channels_last=False, **kwargs):
     Returns:
         dict with the centers of sources detected by SEP detection algorithm.
     """
-    if isinstance(batch["blend_images"], dict):
+    if "blend_images" not in batch.keys():
```
ismael-mendoza, line 107: Same here (see above?)

tests/test_measure.py
```diff
@@ -90,13 +88,14 @@ def test_measure_kwargs():
...
+    print(results["catalog"])
     x_peak, y_peak = (
-        results["catalog"][0]["x_peak"].item(),
-        results["catalog"][0]["y_peak"].item(),
+        results["catalog"]["sep_measure0"][0]["x_peak"].item(),
+        results["catalog"]["sep_measure0"][0]["y_peak"].item(),
     )
-    detected_centers = np.array([[x_peak, y_peak]])
-    target_detection = np.array([[65.495, 51.012]])
+    print(x_peak)
```
ismael-mendoza, line 96:
```suggestion
```

tests/test_measure.py
```diff
@@ -90,13 +88,14 @@ def test_measure_kwargs():
         btk.measure.sep_measure, measure_kwargs=[{"sigma_noise": 2.0}, {"sigma_noise": 3.0}]
     )
     _, results = next(meas_generator)
-    results = results["sep_measure0"]  # extract single element from dict.
+    print(results["catalog"])
```
ismael-mendoza, line 91: Remove extraneous print statements - 
```suggestion
```

ismael-mendoza (approved):
It looks basically ready to merge @thuiop

btk/measure.py
ismael-mendoza: What was wrong with the above approach? (was it too cryptic?)

btk/draw_blends.py
thuiop: Yes.

btk/measure.py
thuiop: You're right, this is my fault ; I changed it when I had reversed the order for the surveys, but I ended up reversing it again so I should put it back as it was before.

ismael-mendoza:
Looks ready to merge, thanks @thuiop

PR #168, opened 2021-06-04
Multiband COSMOS galaxies

mpaillassa:
The aim of this PR is to add the possibility to use multiband magnitude information when available to compute more realistic fluxes for COSMOS galaxies. The idea is simply to have them in the provided COSMOS catalog, hence in the BTK table and to use them in `render_single`.

In order to have a better view of the COSMOS catalogs, here are some small explanations (I will make a proper documentation here or in another PR later).
Basically, BTK draws COSMOS galaxies through galsim using `galsim.COSMOSCatalog` and its `makeGalaxy` method. The instantiation of `galsim.COSMOSCatalog` requires two catalogs:
* One containing information necessary to draw 'real' galaxies (galaxy stamps, PSF stamps and noise characteristics)
* One containing information about parameter fits to the galaxies (sersic parameters, bulge-to-disk ratios, etc)

While both catalogs are required to instantiate `galsim.COSMOSCatalog`, the second one is necessary only if one wants to draw 'parametric' galaxies. But in BTK we draw 'real' galaxies that need only the information of the first catalog. 
Still, the second catalog can be used to retrieve some galaxy parameters within BTK, and BTK actually uses it to get `flux_radius` and a size estimation of the galaxies. So if one wants to provide new COSMOS catalogs, the second catalog needs to contain at least the 'flux_radius' column but it does not need to contain all the other fit parameters that the official COSMOS data set does.

So finally, about the changes in this PR, our plan for multiband images is to provide multiband magnitudes through the second catalog. BTK simply checks if the columns are here, use them if so, or reverse to the current way otherwise (i.e. using the same magnitude for all bands). The format of the columns that are looked for is: <filt.name>_<survey.name> so if the user provides columns in such format they can be used for any survey and filter.


btk/draw_blends.py
```diff
@@ -456,7 +456,13 @@ class CosmosGenerator(DrawBlendsGenerator):
     def render_single(self, entry, filt, psf, survey, extra_data):
         """Returns the Galsim Image of an isolated galaxy."""
         galsim_catalog = self.catalog.get_galsim_catalog()
-        gal_flux = get_flux(entry["ref_mag"], filt, survey)
+
+        # get galaxy flux
+        try:
+            gal_flux = get_flux(entry["_".join([filt.name, survey.name])], filt, survey)
```
thuiop, line 462:
```suggestion
            gal_flux = get_flux(entry[f"{filt.name}_{survey.name}"], filt, survey)
```
is probably more readable. Also, wouldn't reversing the survey and the filter make more sense ?

ismael-mendoza: I agree also with reversing the order of filter and survey (since filters live inside surveys)

mpaillassa: Yes indeed I just took back the column name from the underlying [EL COSMOS](http://cesam.lam.fr/aspic/files/elcosmos/readme_elcosmos.html) catalog but we can reverse.

mpaillassa:
Should I go for the documentation about Cosmos catalogs in this PR or in another one ?

ismael-mendoza:
Hi @mpaillassa - thanks for opening this PR! 

Quick question, where you thinking of adding an example catalog that has the new flux entries as well ? 

> Should I go for the documentation about Cosmos catalogs in this PR or in another one ?

It would be great if you add that in this PR too, since the changes look pretty minimal otherwise

mpaillassa:
Yes I plan to make a similar subset than `data/cosmos/real_galaxy*` so that we can show this in a tutorial.

(I was also thinking about making a new section in the advanced features tutorial about the custom PSF feature (and eventually adding some real PSF images on the repository) but this would be for another PR)

ismael-mendoza:
> Yes I plan to make a similar subset than data/cosmos/real_galaxy* so that we can show this in a tutorial.

Thanks! 

> (I was also thinking about making a new section in the advanced features tutorial about the custom PSF feature (and eventually adding some real PSF images on the repository) but this would be for another PR)

Yes, that sounds great :)

mpaillassa:
I just committed some changes that include:
* Removing of a Gaussian smoothing step in the COSMOS galaxy rendering. We discussed with @herjy about this and found it is not necessary, especially if we use a large PSF to reconvolve the galaxies.
* Adding the example catalog and stamps of our Galsim COSMOS data set extension. 
* Swapping the y and z filter sky brightenesses of HSC. I just realized that those values were reversed when double checking at [this page](https://www.naoj.org/Observing/Instruments/SCam/exptime.html) (the order in BTK and in the table are not the same). This change may need to be elsewhere now as the CLI changes changed `survey.py`

mpaillassa:
I have started rebasing the branch to main. But I realized that I have run one rebase while my main branch was not up to date so the rebasing is not complete to the last main commits.
When I rebase again after having updated my main branch, I have one conflict in `survey.py`, which is expected, but when I solve it and do `git rebase --continue` it does not work, telling me `No changes - did you forget to use 'git add'?`. Any clue bout this ? (It is a bit frustrating, I should have probably merged instead of rebasing...)

mpaillassa:
Ok I have been able to rebase. I have swapped the wrong sky brightness y and z HSC values in the config file.
I have also added a test function to run the new catalogs I have uploaded.

I think it is ready to be merged. If I remember correctly what we talked about the plan would now be to merge this PR and to modify the COSMOS notebook in the #179 PR, right ?

btk/draw_blends.py
```diff
@@ -467,7 +473,7 @@ def render_single(self, entry, filt, psf, survey, extra_data):
         # around them, so a gaussian kernel convolution smoothes it out.
         # It has the slight disadvantage of adding some band-limitedness to the image,
         # but with a small kernel, it's better than doing nothing.
-        gal = galsim.Convolve(gal, galsim.Gaussian(sigma=2 * survey.pixel_scale))
+        # gal = galsim.Convolve(gal, galsim.Gaussian(sigma=2 * survey.pixel_scale))
```
ismael-mendoza, line 476: Can we remove this commented out line? (and the comment from Remy above?) Can you remind me why we needed this?

ismael-mendoza (approved):
Hi @mpaillassa , looks good! Just one small comment below

btk/draw_blends.py
mpaillassa: Oh yes I left it in case it would be useful for one who wants to simulate images with sharp PSFs (this was the first intention of Remy when he added this). But maybe we can just remember it and remove it from the code.

ismael-mendoza: Thanks for the quick reply! Is there a good definition for 'sharp' ? Ideally we can check if the PSF provided by the user is sharp and then apply this in that case. But perhaps it's best if we remove this for now, open a separate issue and resolve this later.

mpaillassa: I think Remy considered sharp in PSF size in the order of the HST PSF. In any case I think it is sure that this smoothing is not necessary for ground-based instrument simulations. Maybe we could discuss this with Remy indeed.

ismael-mendoza: Sounds good just opened #181

ismael-mendoza:
> I think it is ready to be merged. If I remember correctly what we talked about the plan would now be to merge this PR and to modify the COSMOS notebook in the #179 PR, right ?

Yes please! I'm about to merge this PR

PR #169, opened 2021-06-07
New scarlet notebook

thuiop:
@ismael-mendoza seems like you force-pushed the branch so the one on main is currently not the good one.


thuiop:
It should be ready to be merged now.

ismael-mendoza:
oops, thanks for letting me know and for fixing it!

PR #178, opened 2021-07-09
Switch order of dictionaries for metrics

thuiop:
Basically the output of the MetricsGenerator was like `results[meas_function][survey]["galaxy_summary"]` and is now `results["galaxy_summary"][meas_function][survey]` so that the order is the same as for the MeasureGenerator output. Please note that the survey layer is only there in the multiresolution case.
I also fixed the save (at least partially) which was not working correctly after the multiresolution PR.


ismael-mendoza (approved):
Looks good, thanks for making this consistent. Feel free to merge when ready

ismael-mendoza:
> I also fixed the save (at least partially) which was not working correctly after the multiresolution PR.

@thuiop what do you mean by 'partially'? Can you let me know maybe via an issue if there is still some part of the save that doesn't work? Thanks!

thuiop:
> 
> 
> > I also fixed the save (at least partially) which was not working correctly after the multiresolution PR.
> 
> @thuiop what do you mean by 'partially'? Can you let me know maybe via an issue if there is still some part of the save that doesn't work? Thanks!

I am not sure that it works exactly as it should ; the tests are passing but I am not a hundred percent sure that everything is correct now and I have to make some more tests.

PR #180, opened 2021-07-15
Changed metrics default value to NaN

### LINKED ISSUE #141: Add a way to select the default value in case of measurement error
Either we want to use NaNs and handle them, or we need a way to choose which value is used throughout the measurements to spot them afterwards and in the plotting methods.

That value is generally referred to as the `fill_value` but another name could be used.

thuiop:
Closes #141 

Instead of having inconsistent values for the metrics when they could not be computed, the value is now NaN. I am not sure if it is necessary to allow for the user to change it.


ismael-mendoza:
Thanks @thuiop !

Issue #88, opened 2021-01-29
Double check survey parameters in survey.py


ismael-mendoza:
Once #87 is merged, could someone help me double check the survey parameters in `survey.py`. I marked the fields I'm especially concerned about with a `#TODO`. Thanks

PR #187, opened 2021-07-20
Survey parameters double check

mpaillassa:
I create this PR in regard of the issue #88. I have checked each survey and matched parameters with the links from where they can be retrieved. Here is a summary of what I have checked so far (especially parameters I could not double check), along with some questions.

General comments:
* I could not find back any airmass value...
* I have double checked all filter zeropoints for HSC, LSST, Euclid and DES thanks to some code from [this WLDeblending issue](https://github.com/LSSTDESC/WeakLensingDeblending/issues/19).
* I realized that we use the same central wavelength values for the same filter of different surveys. But I am not sure that for instance all the r filters of different surveys have the same central wavelength so this might be an issue ?

CFHT:
* The link to compute zeropoint is dead and I could not find it back. I found [this](http://www.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/en/megapipe/docs/filt.html) but I am not sure it can help. 
* I did not get where to find extinction values and effective area.
* If we solve the two points above we could also add the other CFHT filters (now there are r and i but there are also u, g and z).

DES:
 * I could not find back the sky brightness value for the Z band.
 
Euclid:
* I have set the extinction to 0 (as for HST) but we are not sure about it (in any case the airmass term is also zero so it does not change the flux computation)

HSC:
* I updated seeing values to dr2 data.
* I did not get where to find extinction values.
* I noticed that the effective area does not take into account the secondary mirror obscuration. I will investigate this.

HST:
* I found some parameters for the [ACS WFC](https://hst-docs.stsci.edu/acsdhb) but the link that was already here is about [WFC3](https://hst-docs.stsci.edu/wfc3ihb) and I am not sure which one we want to have in BTK ?
* I think that the mirror diameter is well 2.4m so we could compute a "naive" effective area without considering obscuration in order to replace the 1 default value for now.

Rubin:
* I have not double checked as I guess it has already been done !


thuiop:
Thanks for all this work !

I believe extinction is for the atmosphere so it should be set to 0 for space-based surveys ; we should probably set the airmass and zeropoint airmass to 0 to avoid confusing users. When in doubt for the airmass value I think we should set both the airmass and the zeropoint airmass to 1.

Your comment about the central wavelength is correct ; though I suspect this will only change the PSF by a tiny bit, and having an uniform PSF over the filter is already an approximation.

conf/surveys/HST.yaml
```diff
@@ -1,8 +1,10 @@
-# Source https://hst-docs.stsci.edu/display/WFC3IHB/6.6+UVIS+Optical+Performance#id-6
+# pixel_scale: https://hst-docs.stsci.edu/acsihb/chapter-3-acs-capabilities-design-and-operations/3-5-acs-quick-reference-guide
+# exp_time: https://hst-docs.stsci.edu/acsdhb/chapter-1-acs-overview/1-2-basic-instrument-operations
+# fwhm: https://hst-docs.stsci.edu/display/WFC3IHB/6.6+UVIS+Optical+Performance#id-6
 # .6UVISOpticalPerformance-6.6.1 800nm
 HST:
   name: "HST"
-  pixel_scale: 0.06
+  pixel_scale: 0.05
```
thuiop, line 7: I think here we should use the parameters for COSMOS specifically (maybe we should rename the survey as HST_COSMOS ?). In that case, the pixel scale is actually 0.03 (they used multidrizzle), and the parameters are from the ACS as you surmised.

conf/surveys/HST.yaml
```diff
@@ -11,7 +13,7 @@ HST:
     f814w:
       name: "f814w"
       sky_brightness: 22
-      exp_time: 3000
+      exp_time: 900
```
thuiop, line 16: Should be 2028 seconds for COSMOS https://arxiv.org/pdf/astro-ph/0703095.pdf

conf/surveys/CFHT.yaml
```diff
@@ -1,28 +1,14 @@
...
+# pixel_scale, sky_brightness: http://www.cfht.hawaii.edu/Instruments/Imaging/Megacam/generalinformation.html
+# exp_time, fwhm: https://www.cfht.hawaii.edu/Science/CFHTLS/T0007/CFHTLS_T0007-TechnicalDocumentation.pdf
+# zeropoint: http://www1.cadc-ccda.hia-iha.nrc-cnrc.gc.ca/community/CFHTLS-SG/docs/extra/filters.html
 CFHT:
   name: "CFHT"
-  pixel_scale: 0.185
+  pixel_scale: 0.187
   effective_area: 8.022
   mirror_diameter: 3.592
   airmass: 1.0 # TODO: double-check
```
thuiop, line 9: Should be 1.2, but I did not find the zeropoint airmass so we should also set that one to 1.2.

conf/surveys/HST.yaml
mpaillassa: Yes this is also what I thought about. If @ismael-mendoza is ok I can change HST to HST_COSMOS and set the corresponding parameters.

conf/surveys/CFHT.yaml
mpaillassa: Ok I changed it. Do you have a reference for this value ?

thuiop: https://www.cfht.hawaii.edu/Instruments/Imaging/Megacam/dietmegacam.html

thuiop:
Actually, I wonder if we should not use this opportunity to change the zeropoints to the usual value instead of the WeakLensingDeblending convention. @ismael-mendoza  what do you think ? I can take some time to do it.

conf/surveys/HST.yaml
ismael-mendoza: yes please go ahead! @mpaillassa

ismael-mendoza:
Thanks @mpaillassa for your work on this PR ! 

@thuiop that sounds good to me, please if you could also update the corresponding documentation that would be great

conf/surveys/HST.yaml
```diff
@@ -1,19 +1,22 @@
-# Source https://hst-docs.stsci.edu/display/WFC3IHB/6.6+UVIS+Optical+Performance#id-6
+# pixel_scale: https://hst-docs.stsci.edu/acsihb/chapter-3-acs-capabilities-design-and-operations/3-5-acs-quick-reference-guide
+# exp_time: https://hst-docs.stsci.edu/acsdhb/chapter-1-acs-overview/1-2-basic-instrument-operations
+# fwhm: https://hst-docs.stsci.edu/display/WFC3IHB/6.6+UVIS+Optical+Performance#id-6
+# zeropoint : https://galsim-developers.github.io/GalSim/_build/html/real_gal.html
 # .6UVISOpticalPerformance-6.6.1 800nm
 HST:
   name: "HST"
```
thuiop, line 7:
```suggestion
  name: "HST_COSMOS"
```

aboucaud:
Coming a bit late on this but this is awesome work @mpaillassa and @thuiop ! 👏🏼 

It makes me wonder whether it would not be of high community value to make that survey part a small external library (without dependencies) so it could be reused by any other project. That way, it would benefit many more people and could also be community-backed in the sense that there are many more astronomers will lay eyes on these numbers and carefully double-check them.

If you like that thought, it'd be happy to contribute to make it happen.
Let me know what you think @ismael-mendoza @thuiop @mpaillassa.

thuiop:
This could be interesting but right now the parameters of the surveys are really "BTK-oriented" so I do not know how much it would be useful. Having a database of parameters for different surveys does sound interesting though.

mpaillassa:
Yes I guess it could be interesting, especially if it reaches people that could double check them as well.

By the way, I checked only the HSC zeropoints for now, but where did you get them ? For instance I found some values at [this link](https://www.naoj.org/Observing/Instruments/HSC/sensitivity.html). Values are really close so it does not change much but I was just curious then.

And now that we changed the zeropoint convention we may want to add the other CFHT filters (at first I did not add them because I did not know how to compute the zeropoints with the old convention)

ismael-mendoza:
I agree with @thuiop and @mpaillassa , I think it could be interesting to have it as an external library so other people can use it / double-check values. But I'm not sure how exactly to put it in a more useful format that is not BTK specific - currently it's in the form of configuration files that are specifically designed to work with the `Survey` / `Filter` classes. @aboucaud if you have any ideas regarding this I think that would be great, I'm also happy to discuss more during the meeting this Wednesday if you can make it

aboucaud:
Sure, let's discuss this on Wednesday. I should be able to make it this time 😃

thuiop:
> Yes I guess it could be interesting, especially if it reaches people that could double check them as well.
> 
> By the way, I checked only the HSC zeropoints for now, but where did you get them ? For instance I found some values [here](https://www.naoj.org/Observing/Instruments/HSC/sensitivity.html). Values are really close so it does not change much but I was just curious then.
> 
> And now that we changed the zeropoint convention we may want to add the other CFHT filters (at first I did not add them because I did not know how to compute the zeropoints with the old convention)
I'm pretty sure I computed them from the ones which were already there. Zeropoints tend to be measured again over time so they may change a little bit.

I agree for the CFHT filters ; I was wondering if maybe only those two were used for a particular weak lensing survey (CFIS ?).

PR #205, opened 2021-07-21
Forgot to update rubin in tutorial

ismael-mendoza:


docs/source/tutorials.rst
```diff
@@ -168,90 +168,90 @@ You may want to define your own survey if you wish to modify some parameters or
   }
 
   Rubin = btk.survey.Survey(
-      "LSST",
+      "Rubin",
```
aboucaud, line 171: I do not understand this change. The observatory is called Vera C. Rubin but the survey is still called LSST.

ismael-mendoza: That is a good point! I'll revert it, thanks @aboucaud :)

PR #215, opened 2021-07-28
Multiresolution flag

### LINKED ISSUE #203: make MR flag more explicit
https://github.com/LSSTDESC/BlendingToolKit/blob/44ddbeefeb20c98ed4ebadc07c367f96648798a7/btk/measure.py#L109

I still think this line might be confusing for anyone trying to implement their own measure function. I suggest we provide some sort of internal flag like `is_mr` that users won't be confused by.

thuiop:
Closes #203


ismael-mendoza (approved):
Perfect! This is exactly what I had in mind

PR #223, opened 2021-09-29
Added checks for catalog-survey compatibility

### LINKED ISSUE #207: Raise nice exceptions when using incompatible catalogs
- [ ]  Check whether a given survey is appropriate for a catalog (potentially can be added in DrawBlendsGenerator as a class attribute)
- [ ]   check certain mandatory column names are present in the given catalog and fail with a nice exception if they are not

thuiop:
Closes #207


ismael-mendoza (approved):
This looks like a simple and great solution. Thanks @thuiop !

tests/test_error_cases.py
```diff
@@ -14,6 +16,16 @@
...
+COSMOS_CATALOG_PATHS = [
+    str(data_dir / "cosmos/real_galaxy_catalog_23.5_example.fits"),
+    str(data_dir / "cosmos/real_galaxy_catalog_23.5_example_fits.fits"),
+]
+
+COSMOS_EXT_CATALOG_PATHS = [
+    str(data_dir / "cosmos/real_galaxy_catalog_26_extension_example.fits"),
+    str(data_dir / "cosmos/real_galaxy_catalog_26_extension_example_fits.fits"),
+]
+
```
aboucaud, line 28: These should be defined in the conftest.py to be used by any test.

Also make sure you separate the `"cosmos/xxx.fits"` into `"cosmos" / "xxx.fits"` so that pathlib can resolve Windows filepath, otherwise there is no point in using pathlib.

PR #226, opened 2021-10-01
Changed random number generation to use the most recent numpy API

### LINKED ISSUE #69: Ensure reproducibility of image production runs

thuiop:
I migrated the BTK to use the newest numpy random number generation API. However, this causes an issue with the multiprocessing ; it no longer returns the same thing as the non multiprocessing case (I am not sure how it managed to do that in the first place actually). It is still reproducible, even for the multiprocessing case (with a set number of CPUs). Hopefully, this should make the simulations completely reproducible as long as the multiprocessing is the same. What are your thoughts on this ?

Closes #69.


ismael-mendoza (approved):
Thanks Thomas this looks great, just a couple of small questions below. 

I think it's ok that using different number of CPUs does not give the same result. As long as you can reproduce results by using the same number of CPUs we should be good. Feel free to remove the corresponding unit test (that's the only one failing right?)

btk/sampling_functions.py
```diff
@@ -190,18 +194,20 @@ class BasicSampling(SamplingFunction):
     Includes magnitude cut, restriction on the shape, shift randomization.
     """
 
-    def __init__(self, max_number=4, stamp_size=24.0, maxshift=None):
+    def __init__(self, max_number=4, stamp_size=24.0, maxshift=None, rng=np.random.default_rng()):
```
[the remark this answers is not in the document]
thuiop: No, the ones with the 0 use the seed 0. They should only be in the tests.

btk/draw_blends.py
```diff
@@ -403,7 +417,7 @@ def render_blend(self, blend_catalog, psf, filt, survey, extra_data):
         if self.add_noise:
             if self.verbose:
                 print("Noise added to blend image")
-            generator = galsim.random.BaseDeviate(seed=np.random.randint(99999999))
+            generator = galsim.random.BaseDeviate(seed=self.rng.integers(100000))
```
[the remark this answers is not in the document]
thuiop: Yes, it will give the same seed to the Galsim one everytime so it should be fine.

btk/sampling_functions.py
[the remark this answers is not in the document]
ismael-mendoza: gotcha

aboucaud:
Thanks @thuiop for taking on this task but IMHO this PR was merged a bit too quickly (<24h) for such an important step. 
Random generation is the fondation of reproducibility so I think some of the choices made here should require more discussions before being merged to `main`. 
I serve mostly as a BTK reviewer so I'm just sharing my opinion. As maintainers, this is your call of course.

thuiop:
You may be right. What comments did you have in mind ?

ismael-mendoza:
Hi @aboucaud - thanks for your comment. We are definitely happy to discuss this more and revert/make changes as necessary. Feel free to let us know additional things we might want to discuss in a github issue or during the meeting on Wednesday. Thanks!

btk/draw_blends.py
```diff
@@ -130,6 +130,7 @@ def __init__(
         indexes=None,
         channels_last=False,
         save_path=None,
+        rng=None,
```
aboucaud, line 133: here should be `seed` and not `rng`

btk/draw_blends.py
```diff
@@ -182,6 +185,17 @@ def __init__(
...
 
+        if rng is None:
+            self.rng = np.random.default_rng()
+        elif isinstance(rng, int):
+            self.rng = np.random.default_rng(rng)
+        else:
+            try:
+                rng.random()
+            except AttributeError:
+                raise AttributeError("The random generator you provided is invalid.")
```
aboucaud, line 196: so here if I give `rng = numpy.random` this will work but good luck for reproducibility.
I guess we should restrict only the cases to
- a seed is provided
- a default seed is used

btk/draw_blends.py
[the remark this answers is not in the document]
aboucaud: For clarity, please separate the logic in two lines 

```python
galsim_seed = self.rng.integers(xxx)
galsim.random.BaseDeviate(seed=galsim_seed)
```

But I am still not convinced that this will always return the same seed. It will depend on the order of the computations. If there is a `if` statement somewhere before that calls to `self.rng` this will return a different value for the seed.

That's why I think we need to be careful here because the way you tested it might not be the only way it will be used.

btk/sampling_functions.py
```diff
@@ -55,18 +55,20 @@ def compatible_catalogs(self):
 class DefaultSampling(SamplingFunction):
     """Default sampling function used for producing blend tables."""
 
-    def __init__(self, max_number=2, stamp_size=24.0, maxshift=None):
+    def __init__(self, max_number=2, stamp_size=24.0, maxshift=None, rng=np.random.default_rng()):
```
aboucaud, line 58: Having `default_rng()` without specified seed as the default value of a keyword does not seem a good idea to me.
I'd rather have `None` and then do something about it.

tests/test_draw.py
```diff
@@ -14,10 +14,10 @@ def get_draw_generator(
     add_noise=True,
     fixed_parameters=False,
     sampling_function=None,
+    rng=np.random.default_rng(0),
```
aboucaud, line 17: here the seed is set to 0 without justification.

tests/test_draw.py
```diff
@@ -48,22 +48,19 @@ def get_draw_generator(
...
     )
     return draw_generator
 
 
 class TestMultiprocessing:
     def test_multiprocessing(self):
-        np.random.seed(0)
         b_size = 16
         cpus = np.min([mp.cpu_count(), 16])
+        rng = np.random.default_rng(0)
```
aboucaud, line 60: Here again seed set to 0

tests/test_draw.py
aboucaud: my bad, I get it, it's for testing.
Still it may be a good idea to have the test seed apparent somewhere

```python
TEST_SEED = 0
...
rng = np.random.default_rng(TEST_SEED)
```

aboucaud:
Here are some of the comments I had.

Concerning the multiprocessing, it might be valuable to have a look at
https://albertcthomas.github.io/good-practices-random-number-generators/

I hope I'll be able to make it to the telecon tomorrow.

btk/draw_blends.py
thuiop: Why ? It is not necessary a seed, it can directly be a `numpy.random.Generator`.

btk/draw_blends.py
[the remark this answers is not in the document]
thuiop: Why would there be an additional `if` statement if we use the same parameters for the generation ?

tests/test_draw.py
thuiop: This can be a good idea.

ismael-mendoza:
Another example from Axel: https://github.com/LSSTDESC/descwl-shear-sims/blob/master/shear_meas_tests/test_shear_meas.py

thuiop:
> Another example from Axel: https://github.com/LSSTDESC/descwl-shear-sims/blob/master/shear_meas_tests/test_shear_meas.py

This one uses RandomState, which is from the old numpy API (which is still working but we should not start using it now).

PR #229, opened 2021-10-06
Update galsim-hub generative model to TF2

### LINKED ISSUE #224: Run Galsim_Hub model in tensorflow 2.0
It appears that having tensorflow 1.0 as a dependency keeps giving us security warnings and comes with other problems (like compatibility with other newer packages). It would be nice if we could migrate the galsim_hub model to tensorflow 2.0 somehow - I think @aboucaud , @thuiop  had some ideas?

aboucaud:
This PR aims at incorporating the ensentials of galsim-hub generative model API while switching to the use of `tensorflow` 2.x

Fixes #224

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Install poetry dependencies
         run: |
           poetry run pip install Cython
-          poetry install --extras "scarlet"
+          poetry install --extras "galsim-hub scarlet"
           poetry run pip install git+https://github.com/pmelchior/scarlet
 
       - name: Run Tests
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -47,7 +47,7 @@ jobs:
       - name: Install poetry dependencies
         run: |
           poetry run pip install Cython
-          poetry install --extras "galsim-hub scarlet"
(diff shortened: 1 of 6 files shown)


thuiop:
Shouldn't we look for integrating this into Galsim Hub directly instead of in BTK ?

ismael-mendoza:
> So do we plan on merging this now, and change it later if we make modifications to galsim_hub directly ?

Yes, although I find it unlikely that we will want to modify galsim_hub directly

ismael-mendoza:
> The coverage check should act as a warning but not be blocking. 👇🏼
For instance it does not take into account that the code is actually tested in the notebooks.

I agree, thanks for the suggestion

ismael-mendoza:
Hi @aboucaud - based on our discussions during the sprint week perhaps the easiest thing would be if you could provide me the model that needs to be downloaded (or should we ask Francois for it? ) and I can find a way to add it to the github repo ? (probably git lfs but not downloading it everytime it's needed)

ismael-mendoza:
Hi @aboucaud and @thuiop , as we discussed in the last meeting. We have decided to delay the implementation of tensorflow/galsim hub after the v1.0.0 release. I will close this PR but keep the branch open, feel free to work more on this whenever you have tim but there is no rush anymore.

PR #242, opened 2021-11-02
Small fixes to the measure file

### LINKED ISSUE #240: small issues with measurement.py
* I think the conversion between pixel and ra/dec is wrong (double check)
* add error checking when measure kwargs is a dictionary and not a list of dictionaries.

thuiop:
Closes #240


ismael-mendoza (approved):
This works, thanks @thuiop. 

I guess it's a little annoying to have to remember the correct units

PR #248, opened 2021-11-26
WIP: Adding support for shear

### LINKED ISSUE #98: Add lensing shear to blends
We want functionality to add lensing shear to blends. We can make use of GalSim implementation. Based on our discussion in our BTK telecon, we need to take care of following things.
- Lensing shear needs to be added for each object, since shear is different for sources at difference redshift. We could specify shear for each object in an input catalog. In this way, people could add, e.g., in a simplest case a constant shear or in a complicated case variable shear computed from N-body simulations.
- We might want to check with Matt Becker things like astrometric shift due to lensing.

thuiop:
Closes #98 . I just added basic support for shearing individual galaxies with a constant shear. This should not be merged before version 1.0. Feel free to post any suggestions you may have.


btk/sampling_functions.py
```diff
@@ -259,6 +259,82 @@ def __call__(self, table, **kwargs):
         return blend_table
 
 
+class DefaultSamplingShear(SamplingFunction):
+    """Default sampling function used for producing blend tables."""
+
+    def __init__(self, max_number=2, stamp_size=24.0, maxshift=None, shear=None, seed=DEFAULT_SEED):
```
ismael-mendoza, line 265: why not make the default (0.0, 0.0) and avoid the if statement?

btk/draw_blends.py
```diff
@@ -631,13 +635,14 @@ def render_mini_batch(self, blend_list, psf, wcs, survey, seed):
...
         base_image = extra_data
-        base_image = galsim.Convolve(base_image, psf)
+        gal_conv = galsim.Convolve(base_image, psf)
         gal_flux = get_flux(entry["ref_mag"], filt, survey)
-        base_image = base_image.withFlux(gal_flux)
-        base_image = base_image.shift(entry["ra"], entry["dec"])
-
+        gal_conv = gal_conv.withFlux(gal_flux)
+        gal_conv = gal_conv.shift(entry["ra"], entry["dec"])
+        if "g1" in entry.keys():
```
ismael-mendoza, line 642: this `if` statement can be avoided too if the default is 0.0 for shear everywhere in the code

btk/sampling_functions.py
```diff
@@ -259,6 +259,82 @@ def __call__(self, table, **kwargs):
         return blend_table
 
 
+class DefaultSamplingShear(SamplingFunction):
```
ismael-mendoza, line 262: Right now this function is basically the same as `DefaultSampling`, I guess it's ok since we want to keep the base one simple for tutorial purposes. 

Another option is to have this new class be a children of DefaultSampling and rewrite the `__call__` to first call the parent class function and then add the shear stuff afterwards. What do you think @thuiop ?

ismael-mendoza (changes requested):
Thanks @thuiop  for starting this! I just made a couple of comments and suggestions below

btk/sampling_functions.py
thuiop: I thought about this but it may be useful if we wish to support something else than constant shear.

btk/sampling_functions.py
thuiop: I guess this could be a good idea, yes.

btk/draw_blends.py
```diff
@@ -494,6 +494,8 @@ def render_single(self, entry, filt, psf, survey, extra_data):
         pix_stamp_size = int(self.stamp_size / survey.pixel_scale)
         try:
             gal = get_catsim_galaxy(entry, filt, survey)
+            if "g1" in entry.keys():
```
ismael-mendoza, line 497:
```suggestion
            if "g1" in entry.keys() and "g2" in entry.keys():
```

btk/draw_blends.py
```diff
@@ -544,6 +546,8 @@ def render_single(self, entry, filt, psf, survey, extra_data):
         gal = galsim_catalog.makeGalaxy(
             entry["btk_index"], gal_type="real", noise_pad_size=0
         ).withFlux(gal_flux)
+        if "g1" in entry.keys():
```
ismael-mendoza, line 549:
```suggestion
        if "g1" in entry.keys() and "g2" in entry.keys():
```

btk/draw_blends.py
```diff
@@ -494,6 +494,8 @@ def render_single(self, entry, filt, psf, survey, extra_data):
         pix_stamp_size = int(self.stamp_size / survey.pixel_scale)
         try:
             gal = get_catsim_galaxy(entry, filt, survey)
+            if "g1" in entry.keys() and "g2" in entry.keys():
+                gal = gal.shear(g1=entry["g1"], g2=entry["g2"])
```
ismael-mendoza, line 498: ensure it happens after the rotation and not before

ismael-mendoza:
the PR got too out of sync, will create a new one

PR #249, opened 2021-12-09
Fixed noise seed identical over a batch

thuiop:
This should fix the issue of the seed being identical over a minibatch, as mentioned in #69 .

### DIFF
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -194,7 +194,7 @@ class DrawBlendsGenerator(ABC):
         self.verbose = verbose
         self.channels_last = channels_last
         self.save_path = save_path
-        self.seed = seed
+        self.seedseq = np.random.SeedSequence(seed)
 
     def check_compatibility(self, survey):
         """Checks that the compatibility between the survey, the catalog and the generator.
@@ -249,12 +249,11 @@ class DrawBlendsGenerator(ABC):
             wcss[s.name] = wcs
 
             input_args = []
-            ss = np.random.SeedSequence(self.seed)
-            seeds_minibatch = ss.spawn(self.batch_size // mini_batch_size + 1)
+            seedseq_minibatch = self.seedseq.spawn(self.batch_size // mini_batch_size + 1)
 
             for i in range(0, self.batch_size, mini_batch_size):
                 cat = copy.deepcopy(blend_cat[i : i + mini_batch_size])
-                input_args.append((cat, psf, wcs, s, seeds_minibatch[i // mini_batch_size]))
+                input_args.append((cat, psf, wcs, s, seedseq_minibatch[i // mini_batch_size]))
 
             # multiprocess and join results
             # ideally, each cpu processes a single mini_batch
@@ -314,7 +313,7 @@ class DrawBlendsGenerator(ABC):
             }
         return output
 
-    def render_mini_batch(self, blend_list, psf, wcs, survey, seed_minibatch, extra_data=None):
+    def render_mini_batch(self, blend_list, psf, wcs, survey, seedseq_minibatch, extra_data=None):
         """Returns isolated and blended images for blend catalogs in blend_list.
 
         Function loops over blend_list and draws blend and isolated images in each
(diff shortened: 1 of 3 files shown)


btk/draw_blends.py
```diff
@@ -356,6 +355,7 @@ def render_mini_batch(self, blend_list, psf, wcs, survey, noise_seed, extra_data
             )
             blend_image_multi = np.zeros((len(survey.filters), pix_stamp_size, pix_stamp_size))
             for b, filt in enumerate(survey.filters):
+                noise_seed = self.rng.integers(MAX_SEED_INT)
```
ismael-mendoza, line 358: Thanks @thuiop for starting this PR. Unfortunately, I think that the way you wrote it will lead to issues with multiprocessing since the different processes will access `self.rng` in non-deterministic order. So that the `noise_seed` used in each process will depend on which one is "faster".

btk/draw_blends.py
```diff
@@ -249,10 +249,12 @@ def __next__(self):
             wcss[s.name] = wcs
 
             input_args = []
+            ss = np.random.SeedSequence(self.seed)
+            seeds_minibatch = ss.spawn(self.batch_size // mini_batch_size + 1)
```
aboucaud, line 253: If I get it right, each time `__next__()` is called, the seed sequence is re-initialised with the same `self.seed`. 
Hence the minibatch seeds are identical every single time.

Is that what you want here ? (genuine question)

btk/draw_blends.py
```diff
@@ -327,6 +329,8 @@ def render_mini_batch(self, blend_list, psf, wcs, survey, noise_seed, extra_data
             psf (list): List of Galsim objects containing the PSF
             wcs (astropy.wcs.WCS): astropy WCS object
             survey (dict): Dictionary containing survey information.
+            seed_minibatch (numpy.random.SeedSequence): Numpy object for generating
+                random seeds (for the noise generation).
```
aboucaud, line 333: The parameter name should reflect on the object type, or encompass it.
You call it `seed_minibatch` so I would expect being able to pass a seed, not a seed sequence.

I have so far no opinion whether it's easier to pass a seed and initialise a seed sequence from it at the beginning of the function or directly pass the seed sequence. But it could also be both e.g.
```python
if isinstance(seed_minibatch, int):
    seed_minibatch = np.random.SeedSequence(seed_minibatch)
```
Does not really help regarding the naming I mentioned at the top though. :)

btk/draw_blends.py
```diff
@@ -355,9 +359,11 @@ def render_mini_batch(self, blend_list, psf, wcs, survey, noise_seed, extra_data
                 (self.max_number, len(survey.filters), pix_stamp_size, pix_stamp_size)
             )
             blend_image_multi = np.zeros((len(survey.filters), pix_stamp_size, pix_stamp_size))
+            seeds_blend = seed_minibatch.spawn(len(survey.filters))
             for b, filt in enumerate(survey.filters):
+                seed_blend = seeds_blend[b]
                 single_band_output = self.render_blend(
-                    blend, psf[b], filt, survey, noise_seed, extra_data[i]
+                    blend, psf[b], filt, survey, seed_blend, extra_data[i]
```
aboucaud, line 366: all these are seed sequences and not seed values (integers), again this is not made clear enough IMO.

btk/draw_blends.py
aboucaud: After seeing a bit more of the code, I guess the only decent option is to rename `seed_xxx` with `seedseq_xxx` or `seed_seq_xxx` because a seed sequence is a very different from a seed.

aboucaud (commented):
The inclusion of the `SeedSequence` is a very bright move. My point with these comments  is to make sure that developers reading the code clearly understand what is expected at each step, and how to use a seed sequence: spawn other sequences or generate random numbers from it.

btk/draw_blends.py
thuiop: Hmm, you're right, I'm not 100% sure that this is what we want. I think it would be better to initialize the `SeedSequence` in the `__init__`. Otherwise we will just get the same noise realization in each batch.

btk/draw_blends.py
thuiop: I agree, I will do that. I think we should always pass the seed sequence up until the point we actually need the integer seed.

ismael-mendoza (approved):
Hi @thuiop ! This looks ready to merge, any final changes?

thuiop:
It should be good now, I am merging.

PR #254, opened 2022-01-13
Separate noises

thuiop:
As we discussed yesterday, I separated the noise from the galaxy flux from the background noise ; it is possible to select which noise you want to be drawn.


btk/draw_blends.py
```diff
@@ -417,12 +421,20 @@ def render_blend(self, blend_catalog, psf, filt, survey, seedseq_blend, extra_da
             _blend_image += single_image
 
         # add noise.
-        if self.add_noise:
+        if self.add_noise in ["galaxy", "all"]:
```
ismael-mendoza, line 424: Doesn't matter too much but these lists you are creating only to check check conditions can be tuples.

ismael-mendoza (approved):
Looks good! Couldn't find anything wrong with the approach. I think you can merge whenever.

ismael-mendoza:
@thuiop  I think you can merge this whenever, are there any parts of the documentation that will need updating besides the docstring?

thuiop:
> @thuiop I think you can merge this whenever, are there any parts of the documentation that will need updating besides the docstring?
I'll look into it and merge afterwards.

thuiop:
Ok, this is finally mergeable.

PR #256, opened 2022-01-20
Parametric cosmos simulations

b-biswas:
It seems that currently, the CosmosGenerator can simulate only with gal_type='real'.
Thus forming "ghosts" among the simulations.

Perhaps it might be a good option to let the user choose between parametric and real galaxies.


notebooks/01a-cosmos_tutorial.ipynb
```diff
@@ -213,9 +215,9 @@
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "btk",
```
thuiop: Be careful, this should not be changed in the git repo.

thuiop (changes requested):
This is a nice addition ; I thought about it at some point but never actually did it. This will allow people to use the COSMOS catalog parametric models easily. It may be worth adding a mention of this in the documentation.

b-biswas:
> This is a nice addition ; I thought about it at some point but never actually did it. This will allow people to use the COSMOS catalog parametric models easily. It may be worth adding a mention of this in the documentation. Also, it seems that the linting is not correct ; before you make further commits, you should run `pre-commit install`, which will run the linting pipeline automatically every time you commit, which ensures that the code stays clean.

Thanks, @thuiop for your remark. I did try to run this but I am getting some strange errors at the fetch stage. 
I could not resolve this so I had to update the PR without pre-commit. 
I see the following error:
Screenshot 2022-02-10 at 16 05 22

notebooks/01a-cosmos_tutorial.ipynb
b-biswas: this seems to be coming from my jupyter kernel... 
Is this also supposed to be fixed by pre-commit?

thuiop: No. You should take care to change it back to Python 3 before committing ; I don't know of a way to go around this.

b-biswas: Ok, I can do it this way then!
I am having a bit of trouble understanding why exactly this is required though... The metadata of the notebook won't change anything right? 
My actual concern is that other than this line, there are a few more that have been changed in the metadata. Do you mean that the entire metadata section should remain the same? or just this line?

aboucaud: Maybe the `databooks` pre-commit hook is the best tool for our needs 
https://github.com/datarootsio/databooks
available in https://pre-commit.com/hooks.html

But it will strip the metadata, meaning that the user will have to choose a kernel when the server is starting.
(that's not necessarily bad)

aboucaud: Let me know what you think @thuiop and @ismael-mendoza

mpaillassa:
Hello, I looked at the changes and I was wondering if there would not be another way to perform the same cut than GalSim on the BTK table. By reusing directly the same code than GalSim with hardcoded cut values like it is currently done, we would need to change them here as well if they change within GalSim (even though I do not expect that they will change).
Instead, we may get the indexes of entries to keep in the BTK table by using the `getOrigIndex()` method on the entries of the `COSMOSCatalog` that we build and that includes the cut. What do you think ?

notebooks/01a-cosmos_tutorial.ipynb
thuiop: This does seem like something which would be useful.

mpaillassa:
> > I looked at the changes and I was wondering if there would not be another way to perform the same cut than GalSim on the BTK table. By reusing directly the same code than GalSim with hardcoded cut values like it is currently done, we would need to change them here if they change within GalSim (even though I do not expect that they will change). Instead, we may get the indexes of entries to keep in the BTK table by using the `getOrigIndex(self, index)` method on the entries of the `COSMOSCatalog` that we build and that includes the cut. What do you think ?
> 
> @mpaillassa thanks for pointing that out, I tried to make the necessary changes. Let me know if it looks ok to you now!

This looks ok ! Maybe we should set `exclusion_level="marginal"` by default for safety and consistency with GalSim. And people may change it at their own risk then.

notebooks/01a-cosmos_tutorial.ipynb
aboucaud: Good point, don't know that. Needs to be checked

aboucaud: Please @b-biswas write an issue about that so it can be checked in a future work.

aboucaud (dismissed):
All changes looking good now.

b-biswas:
> > > I looked at the changes and I was wondering if there would not be another way to perform the same cut than GalSim on the BTK table. By reusing directly the same code than GalSim with hardcoded cut values like it is currently done, we would need to change them here if they change within GalSim (even though I do not expect that they will change). Instead, we may get the indexes of entries to keep in the BTK table by using the `getOrigIndex(self, index)` method on the entries of the `COSMOSCatalog` that we build and that includes the cut. What do you think ?
> > 
> > 
> > @mpaillassa thanks for pointing that out, I tried to make the necessary changes. Let me know if it looks ok to you now!
> 
> This looks ok ! Maybe we should set `exclusion_level="marginal"` by default for safety and consistency with GalSim. And people may change it at their own risk then.

I changed the default exclusion_level to `marginal` but for that, I had to modify the unit tests because unit tests work only at the exclusion_level `none`. [to use `marginal` cut, the `_selection.fits` file must also be present]
does it look ok now?

mpaillassa:
> I changed the default exclusion_level to `marginal` but for that, I had to modify the unit tests because unit tests work only at the exclusion_level `none`. [to use `marginal` cut, the `_selection.fits` file must also be present] does it look ok now?

Ok, the current state works but we may rather add the `_selection.fits` file to the repo to make the tests with default paremeters ?

b-biswas:
> Ok, the current state works but we may rather add the `_selection.fits` file to the repo to make the tests with default paremeters ?

I was trying to do that actually...
But I need some clarification here. Would it work if I simply sample 200 events from the catalog loaded as astropy table and save it back as `.fits` files?
All events in the small sample currently used for testing survive the  `marginal` cut. So I am not sure if it is useful to have the same set of objects. I will have to create a new sample.

mpaillassa:
> > Ok, the current state works but we may rather add the `_selection.fits` file to the repo to make the tests with default paremeters ?
> 
> I was trying to do that actually... But I need some clarification here. Would it work if I simply sample 200 events from the catalog loaded as astropy table and save it back as `.fits` files? All events in the small sample currently used for testing survive the `marginal` cut. So I am not sure if it is useful to have the same set of objects. I will have to create a new sample.

Is it that the selection file `real_galaxy_catalog_25.2_selection.fits` only works to manage the cut with the full catalog `real_galaxy_catalog_25.2.fits` but that it is not usable with the `real_galaxy_catalog_23.5_example.fits` subsample we have in BTK ?
Indeed I just looked into it and while `exclusion_level="marginal"` is the default in GalSim, it is to be used only with the full 25.2 catalog... But as we only have the 23.5_example in BTK, maybe it would be better to revert to `exclusion_level="none"` as default in BTK so that we can make the tests and notebooks with default parameters. And we let the user manage `exclusion_level` when using the whole data set ? (sorry for changing again)

b-biswas:
> Is it that the selection file `real_galaxy_catalog_25.2_selection.fits` only works to manage the cut with the full catalog `real_galaxy_catalog_25.2.fits` but that it is not usable with the `real_galaxy_catalog_23.5_example.fits` subsample we have in BTK ? Indeed I just looked into it and while `exclusion_level="marginal"` is the default in GalSim, it is to be used only with the full 25.2 catalog... But as we only have the 23.5_example in BTK, maybe it would be better to revert to `exclusion_level="none"` as default in BTK so that we can make the tests and notebooks with default parameters. And we let the user manage `exclusion_level` when using the whole data set ? (sorry for changing again)

I just added the _selection.fits files for the 23.5_example. (just didn't know where to find it)
`marginal` should work for 23.5 as well because in the `apply_exculsion` function I see cuts dedicated to 23.5.
The original tests should now work but the `marginal` cut does not remove any object from the sample repo (all the galaxies survive the cuts). So I am not convinced if it is a good test.

mpaillassa:
> I just added the _selection.fits files for the 23.5_example. (just didn't know where to find it) `marginal` should work for 23.5 as well because in the `apply_exculsion` function I see cuts dedicated to 23.5. The original tests should now work but the `marginal` cut does not remove any object from the sample repo (all the galaxies survive the cuts). So I am not convinced if it is a good test.

Oh yes nice ! I guess this file is available in the `COSMOS_23.5_training_sample.tar.gz` archive, but I actually never downloaded it as I have only used the 25.2 one. Anyway, I guess it should be ok now !

b-biswas:
> Oh yes nice ! I guess this file is available in the `COSMOS_23.5_training_sample.tar.gz` archive, but I actually never downloaded it as I have only used the 25.2 one. Anyway, I guess it should be ok now ?
> 
> Edit: Hm indeed the test does not really cover the case when galaxies are actually removed from the catalog..

Except for the final point, everything seems to be ok for me!

b-biswas:
@aboucaud I am not sure if the last commit was required... 
I added it because the function accepts a new parameter. But the tests were working even without it.

aboucaud (approved):

thuiop:
This seems good to me, but I am a bit confused ; what does the _selection file do exactly ? Does it contain which stamps and fits are good or bad ?

b-biswas:
> This seems good to me, but I am a bit confused ; what does the _selection file do exactly ? Does it contain which stamps and fits are good or bad ?

Yes.
Galsim internally looks for the _selection.fits file (so we don't need to feed the path) in the repo where real and parametric data is stored. Then it places cuts on some of the columns. 
To be specific, the `marginal` and `bad_stamp` exclusion level cuts require the `_selection.fits` file.

ismael-mendoza (approved):
Looks good to me just a couple of small suggestions below @b-biswas. Is there any hold-up on merging that I can help with?

btk/draw_blends.py
```diff
@@ -507,7 +507,68 @@ def render_single(self, entry, filt, psf, survey, extra_data):
...
+                        Used mostly for internal testing purposes.
+            channels_last (bool): Whether to return images as numpy arrays with the channel
+                                (band) dimension as the last dimension or before the pixels
+                                dimensions (default).
+            save_path (str): Path to a directory where results will be saved. If left
+                            as None, results will not be saved.
+            seed (int): Integer seed for reproducible random noise realizations.
+            gal_type (str): string to specify the type of galaxy simulations.
+                            Either "real" (default) or "parametric".
+        """
```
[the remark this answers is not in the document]
b-biswas: galsim does give a relevant error so I did not check it over here

btk/catalog.py
```diff
@@ -108,19 +108,52 @@ def __init__(self, raw_catalog, galsim_catalog, verbose=False):
...
+                [default: "marginal"]
             verbose: whether to print verbose info.
         """
+        assert exclusion_level in (
+            "none",
+            "bad_stamp",
+            "bad_fits",
+            "marginal",
+        ), 'The options for \
+            exclusion_level are: "none", "bad_stamp", "bad_fits", "marginal"'
```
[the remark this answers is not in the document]
b-biswas: Indeed, galsim takes care of the error. 
I just removed these lines!

b-biswas:
@ismael-mendoza, turns out that there bug with the `add_noise` parameter here that is slipping through without any checks. 
I tried to fix it in PR #264. We need to merge that before merging this PR.

b-biswas:
@thuiop maybe we can merge this PR if this looks ok to you?

thuiop:
Yes, it seems good to me !

PR #258, opened 2022-02-09
Add progress bar

thuiop:
I thought it would be nice to have some sort of progress bar when generating the blends ; this should mostly do the trick. There are two main issues : 
- the progress bar only tracks the progress in the loop of the run_minibatch function ; the remainder of the generator takes a bit more time to process
- when using the multiprocessing feature, this will display one bar for each CPU instead of a single bar ; there is sadly no easy way around this

This uses the tqdm package, which may need to be added to the list of dependencies (although it might already be installed anyway as a dependency of another package).


thuiop:
This PR should be ready ; I am not really sure why the test if failing however, it sounds like it is not related to the PR in particular...

ismael-mendoza:
sorry for the delay, I'll look into this now

ismael-mendoza:
Thanks @thuiop I think the progress bar is a good idea and will help people see how much time their simulations will take (if they take a while that is useful to know!) 

> the progress bar only tracks the progress in the loop of the run_minibatch function ; the remainder of the generator takes a bit more time to process

That's fine, I think the `run_minibatch` function should be most of the runtime anyways. I added a description "Generating blends:" so it's clear what the loop is doing. 

> when using the multiprocessing feature, this will display one bar for each CPU instead of a single bar ; there is sadly no easy way around this

Good point, I thought of two ways to "resolve" this: (1) Just do tqdm in the first CPU since hopefully the runtime is around the same for all CPUs (2) Add in the description of the progress bar which CPU is beign used. Do any of you sound good to you @thuiop ?

thuiop:
The second solution sounds good. 
It also looks like I need to rerun the notebooks since the output changed.

ismael-mendoza:
I reran `nbval` in my terminal and everything is passing so I'm merging now. @thuiop let me know if you have additional suggestions and I can incorporate in another PR.

thuiop:
Seems good to me, thanks !

PR #260, opened 2022-02-10
fetch flake8 from github

b-biswas:
I have been facing some issues running pre-commit because there was a problem trying to fetch from gitlab although the server seems to be up. 
So I am not sure if this is a generic issue with HPC systems or something specific to cc-in2p3 but switching to github solved my issue.


aboucaud (approved):

b-biswas:
@herjy isn't there an extra " at the end of line 27 over here [https://github.com/pmelchior/scarlet/pull/262/files](https://github.com/pmelchior/scarlet/pull/262/files)?

PR #264, opened 2022-02-17
fixing problem with add_noise

b-biswas:
This is a continuation of #254 
I updated the config file and the tests for the new options of add_noise


ismael-mendoza (approved):
Good catch! Thanks @b-biswas !

PR #266, opened 2022-02-21
keep notebook output but remove metadata with nbstripout

### LINKED ISSUE #262: Handle notebook metadata
Is there a way to automatically manage the notebook metadata so that CI can also run smoothly?

ismael-mendoza:
Closes #262 

What do you think @b-biswas ? In the future, everyone should just re-run `pre-commit install` and will be all set with the notebook metadata. I don't think this should affect the CI running notebooks.

- [x] add nbstripout to pyproject.toml
- [x] nbstripout current notebooks with flag `--keep-output` 
- [x] add to pre-commit


thuiop (approved):
I like this solution, this seems pretty simple and efficient.

b-biswas:
@ismael-mendoza that's a good solution!

ismael-mendoza:
Thank you both!

PR #284, opened 2022-03-27
Cleanup deps

ismael-mendoza:


thuiop:
Shouldn't this be merged into dev rather than main ?

ismael-mendoza:
> Shouldn't this be merged into dev rather than main ?

good catch thanks!

ismael-mendoza:
Superseded by #288

PR #285, opened 2022-03-29
Fix galcheat in BTK and separate PSF from filter/surveys

ismael-mendoza:
After discussing with @aboucaud [here](https://github.com/aboucaud/galcheat/issues/108), I did some more thinking and  realized that perhaps making the galcheat classes extendable in BTK is not the highest priority. It definitely adds some complexity on both sides (BTK and galcheat) and I remember from a previous BTK meeting that allowing users to change survey configurations is not so useful at the moment. 

In this PR I propose to separate the PSF from survey/filter objects which actually simplifies the code greatly since we can simply use the galcheat classes as is. I think this is also more user friendly and less confusing. Separating out the PSF was actually really simple and now there is less to check (unless I missed something?) 

I'm open to suggestions or comments on this @aboucaud / @thuiop

### DIFF
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -6,10 +6,10 @@ from abc import abstractmethod
 from collections.abc import Iterable
 from itertools import chain
 
+import galcheat
 import galsim
 import numpy as np
 from astropy.table import Column
-from galcheat.survey import Survey
 from galcheat.utilities import mag2counts
 from galcheat.utilities import mean_sky_level
 from tqdm.auto import tqdm
@@ -18,6 +18,7 @@ from btk import DEFAULT_SEED
 from btk.create_blend_generator import BlendGenerator
 from btk.multiprocess import get_current_process
 from btk.multiprocess import multiprocess
+from btk.survey import get_default_psf_with_galcheat_info
 from btk.survey import make_wcs
 
 MAX_SEED_INT = 1_000_000_000
@@ -125,7 +126,7 @@ class DrawBlendsGenerator(ABC):
         self,
         catalog,
         sampling_function,
-        surveys: list,
+        surveys,
         batch_size=8,
         stamp_size=24,
         cpus=1,
@@ -136,6 +137,7 @@ class DrawBlendsGenerator(ABC):
         channels_last=False,
         save_path=None,
         seed=DEFAULT_SEED,
+        psf_func=get_default_psf_with_galcheat_info,
     ):
         """Initializes the DrawBlendsGenerator class.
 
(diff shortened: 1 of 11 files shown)


btk/draw_blends.py
```diff
@@ -176,23 +179,17 @@ def __init__(
...
-            )
+        # get surveys from galcheat.
+        if not isinstance(surveys, Iterable):
+            surveys = [surveys]
+
+        self.surveys = []
+        for s in surveys:
+            if not isinstance(s, str):
+                raise TypeError(f"One of the surveys specified is not a string but type {type(s)}")
+            survey = galcheat.get_suvey(s)
```
aboucaud, line 190: typo
```suggestion
            survey = galcheat.get_survey(s)
```

aboucaud: One side comment here, `get_survey` can raise a ValueError when the survey name does not exist in galcheat.
I made the error message informative though. But so you know if you want to handle that case.

aboucaud (dismissed):
Looks nice, well played :)

btk/draw_blends.py
```diff
@@ -136,15 +137,15 @@ def __init__(
...
     ):
         """Initializes the DrawBlendsGenerator class.
 
         Args:
             catalog (btk.catalog.Catalog): BTK catalog object from which galaxies are taken.
             sampling_function (btk.sampling_function.SamplingFunction): BTK sampling
                 function to use.
-            surveys (list or galcheat.survey.Survey): List of galcheat Survey objects or
-                single galcheat Survey object.
+            surveys (list or list of str): List of galcheat survey names or a single survey name.
```
thuiop, line 148:
```suggestion
            surveys (str or list of str): List of galcheat survey names or a single survey name.
```

btk/draw_blends.py
```diff
@@ -176,23 +179,17 @@ def __init__(
...
-                f"surveys must be a Survey object or an Iterable of Survey objects,"
-                f"but surveys is type {type(surveys)}"
-            )
+        # get surveys from galcheat.
+        if not isinstance(surveys, Iterable):
+            surveys = [surveys]
+
+        self.surveys = []
+        for s in surveys:
+            if not isinstance(s, str):
```
thuiop, line 188: If the PSF is now independent on the survey, shouldn't we still support giving galcheat surveys directly for experienced users ? This should not add too much complexity, I believe.

thuiop (commented):
I am not totally sure how I feel about this, but it does look like it simplifies the code quite a bit.

btk/draw_blends.py
ismael-mendoza: sure I can change that

ismael-mendoza: done !

PR #288, opened 2022-04-07
Cleanup dependencies

ismael-mendoza:
The pandas dependency makes it difficult to support python 3.7 so I removed support of python 3.7 for now.


ismael-mendoza:
The tests are failing but installation works, will merge this into develop and fix tests in a galcheat specific PR. The tests are failing because we need to extend the galcheat survey/filters in BTK.

PR #289, opened 2022-04-07
Fix galcheat by inheritance in BTK

ismael-mendoza:


ismael-mendoza:
Overall a simple change, @aboucaud the only hacky part is the function `_get_btk_survey_from_galcheat_survey` which perhaps you had something else in mind ? Thanks! 

As always any comments @thuiop , @aboucaud would be appreciated!

ismael-mendoza:
Some values were updated in the rubin dataset of galcheat, so tests needed to be updated to reflect that

ismael-mendoza:
@thuiop as you are writing documentation for this you will probably want to mention how to modify survey or how users would create their own survey. 

For modifying surveys or filters the following can be done

```
s = btk.get_surveys('LSST') 
u = s.get_filter('u')

s.pixel_scale = 0.2
u.exposure_time = 1.0
```

Creating a survey/fitler from scratch is more difficult since  it requires passing astropy quantities with units to the galcheat data class constructor. @aboucaud any suggestions on this or were you envisioning creating surveys solely from 
yaml files?

btk/survey.py
```diff
@@ -10,7 +10,57 @@
 import galsim
 import numpy as np
 from astropy.io import fits
-from galcheat.survey import Survey
+
+
+class Survey(galcheat.survey.Survey):
+    """Survey object that extends Galcheat surveys to allow modification."""
```
ismael-mendoza, line 16: add method `from_galcheat_survey()` instead of private method

ismael-mendoza:
Hi @aboucaud - I just added your latest suggestion. Please let me know what you think if you have any others. Otherwise feel free to approve and I'll merge afterwards. Thank you!

tests/test_error_cases.py
```diff
@@ -124,7 +124,7 @@ def compatible_catalogs(self):
             cpus=cpus,
             add_noise=add_noise,
         )
-        draw_output = next(draw_generator)  # noqa: F841
+        _ = next(draw_generator)  # noqa: F841
```
aboucaud, line 127: Is there a reason why you don't just call `next(draw_generator)` ?

aboucaud (dismissed):

tests/test_error_cases.py
ismael-mendoza: good catch!

Issue #293, opened 2022-04-15
General refactoring issue


ismael-mendoza:
In a lot of ways the BTK library has matured enough to be released, and the API for the user is very clear. We have very comprehensive documentations and tutorials which is awesome. However, I still feel like there are some shortcomings in the code structure that make it hard to maintain, extend, and invite new contributors to the library to work on a subset of the code. This will become more important as both the CL and PZ groups are looking to incorporate blending effects into their codes.

I have been thinking about this a lot in the past few weeks and I wanted to write some of my thoughts down here in case other people have ideas too. I think we should still write the paper before this and make some `0.9.X.Y` release as well. Our plans should stay the same but it's good to think ahead a little in terms of what could benefit the community the most.

# Brainstorm

## Modularization

- Currently BTK is designed around a couple of generators `DrawBlendGenerator`, `MeasureGenerator`, `MetricsGenerator`, `BlendGenerator` which are passed one after the other and are coupled together. This makes everything sequential and makes it difficult for the user to extract a particular functionality of BTK.
-  For example, a user who wants to generate images of galaxy blends need to fetch the sampling function and create a separate BTK catalog class which feels burdensome. The same thing happens with the metrics module which is still hard to use as a standalone as pointed out in #286.
- In general, I think we should avoid using Public classes when possible since that adds a layer of mental burden to the user. Example: 
`images = get_blend_images(catalog, survey, sample='uniform')` is a lot easier to understand than  `next(draw_generator)`. Even if they makes sense to us the user might be confused by them and just want a getter function. (this is jut an example, not immediately suggesting we get rid of the Blend Generators)

## Galaxy Model 
- Currently our code for drawing single galaxies is together with the code creating the blends. I think we should have a separate module for individual galaxies. Is this possible? 

## Measurement
-  I keep wondering if `measure.py` is necessary are at all or if it just makes the whole pipeline more difficult to maintain. If our simulation has clear outputs in the form of images and a truth catalog, the user should know how to run their detection / deblending algorithms. 
- We could still have a couple of functions in `measure.py` that run e.g. sextractor and scarlet. But this would really just be barebones implementations to test that our outputs are sensible. I don't think we need a Measure class like we have now.

## Galcheat / Surveys

[...]

PR #296, opened 2022-04-22
Remove cli temporarily

### LINKED ISSUE #295: remove CLI temporarily and add back later for future release

ismael-mendoza:
As discussed, may revive later. I think the code needs to mature more with some suggestions in #293 before a CLI can be implemented. Keeping now means we need to maintain a CLI and the codebase as the codebase evolves which can get complicated. Also, I suspect most people are ok with the python implementation of the code.

Closes #295


thuiop (approved):
Looks good to me.

PR #302, opened 2022-05-09
Release/1.0.0a1

ismael-mendoza:
[Do not merge this branch]

This is just a temporary branch design to track what will get merged into `main` and go into the `1.0.0a1` pre-release of BTK.  (since I did not start doing the develop branch workflow until recently, there are other things in the release beyond what is in the diff but at least it gives us a sense) Most of the changes below have to do with galcheat since that is when I switched to the new workflow

@aboucaud @thuiop @mpaillassa please let me know what you think.  No extensive review necessary just a skim since we already reviewed all these PRs. I'm thinking of doing a 1.0.0 pre-release for now and later do a final release after the internal review. Thanks!


ismael-mendoza:
I went ahead in merged since it doesn't matter too much for the pre-release to be perfect. Still let me know if I missed anything and I can add it in the next pre-release

PR #320, opened 2022-06-16
Cosmos multiband notebook

mpaillassa:
Here is the PR related to #316.

I just added some multiband HSC images using the catalog already there. The next iterations may provide some better looking galaxies but it may be better to stay on the first one so that the user gets the same results.


mpaillassa:
Oh and I just noticed that the plotting function writes "gri images" no matter what bands we select, so we may want to fix this. Maybe in another PR ?

ismael-mendoza:
> Oh and I just noticed that the plotting function writes "gri images" no matter what bands we select, so we may want to fix this. Maybe in another PR ?

@thuiop does #323 address this ?

notebooks/01a-cosmos_tutorial.ipynb
```diff
@@ -68,9 +68,7 @@
    "id": "975ddc32",
```
ismael-mendoza, line 68: <p>I noticed the names of the files are a bit confusing <code>_fits.fits</code>,is there another name we can use to make it clear the information that each file contains?   </p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/LSSTDESC/BlendingToolKit/pull/320/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='8'/>

ismael-mendoza: maybe a different suffix we can use for both catalog files? I think we can change all filenames accordingly if there is a better name you can think of

mpaillassa: Yes I used those names to follow the convention of the original COSMOS dataset, i.e., `catalog.fits` for the catalog containing information for real galaxies and `catalog_fits.fits` for the catalog containing the information for parametric galaxies. The thing is that galsim automatically looks for files named with this convention, so I do not think we can set other names unless we change galsim...
When making our extended galsim COSMOS dataset, I also wondered in which catalog I should put the EL-COSMOS HSC multiband magnitudes (both options are possible because we merge both catalogs in the BTK catalog class) and I think the second one makes sense because those magnitudes come from fits, but I may put them in the first catalog (along with the original HST magnitude of galaxies) if we think it makes more sense.

PR #338, opened 2022-07-19
Sep measure function reworked

atorchylo:
Reworked function `sep_measure` into two functions `sep_singleband_measure` and `sep_multiband_measure`

- `sep_singleband_measure` — a function that performs, detection, segmentation, and deblending running source extractor on a single band (`meas_band_num=3` by default, which is `i `channel if we have `u,g,r,i,z,y` input). It’s basically the same `sep_measure` function but instead of averaging across the channels, we pass a single channel as an input to the Source Extractor.

- `sep_multiband_measure` — a function that performs only detection using all bands of the input image. To do that, I run Source Extractor on each band individually and then merge predictions into one big catalog. To make sure that I don’t add the same source twice, I run a KdTree algorithm and use a threshold distance of `1 arcsec`.

The tests were changed accordingly to the new added funcitonality.


btk/measure.py
```diff
@@ -135,18 +137,107 @@ def basic_measure(
...
+    matching_threshold=1.0,
     sigma_noise=1.5,
     is_multiresolution=False,
     **kwargs,
 ):
-    """Return detection, segmentation and deblending information with SEP.
+    """Returns centers detected with source extractor by combining predictions in different bands.
+
+    NOTE: If this function is used with the multiresolution feature,
+    measurements will be carried on the first survey.
```
ismael-mendoza, line 153: This is somethign about this function that is somewhat implicit and may surprise users, please let me know if you have any ideas how to improve this or if you think this is OK ?

btk/measure.py
```diff
@@ -135,18 +137,107 @@ def basic_measure(
...
+        ra_detections *= 3600
+        dec_detections *= 3600
+
+        # convert to sky coordinates
+        c1 = SkyCoord(ra=ra_detections * units.arcsec, dec=dec_detections * units.arcsec)
+        c2 = SkyCoord(ra=ra_coordinates * units.arcsec, dec=dec_coordinates * units.arcsec)
+
+        # add new coordinates
+        if len(c1) > 0 and len(c2) > 0:
+            idx, d2d, d3d = c1.match_to_catalog_sky(c2)
```
ismael-mendoza, line 208: what exactly are the `d2d` and `d3d` ?

btk/measure.py
```diff
@@ -135,18 +137,107 @@ def basic_measure(
...
     idx,
     channels_last=False,
     surveys=None,
+    matching_threshold=1.0,
     sigma_noise=1.5,
     is_multiresolution=False,
     **kwargs,
 ):
-    """Return detection, segmentation and deblending information with SEP.
+    """Returns centers detected with source extractor by combining predictions in different bands.
```
ismael-mendoza, line 150: Could you add an explanation to the docs explaning exactly how you 'combine' the predictions in different bands?

btk/measure.py
```diff
@@ -135,18 +137,107 @@ def basic_measure(
...
+        if len(c1) > 0 and len(c2) > 0:
+            idx, d2d, d3d = c1.match_to_catalog_sky(c2)
+            d2d = d2d.arcsec
+
+            ra_coordinates = np.concatenate(
+                [ra_coordinates, ra_detections[d2d > matching_threshold]]
+            )
+            dec_coordinates = np.concatenate(
+                [dec_coordinates, dec_detections[d2d > matching_threshold]]
+            )
```
ismael-mendoza, line 216: just to make sure I understand, `ra_coordinates` are the detections from first band, and `ra_detections` on a separate band. You will never add the same center twice because you have the `matching_treshold` right?

ismael-mendoza: This looks correct to me, do you think it's worth checking with a unittest?

tests/test_measure.py
```diff
@@ -3,21 +3,42 @@
...
 def get_meas_results(meas_function, cpus=1, measure_kwargs=None):
-    """Returns draw generator with group sampling function"""
-
+    """Runs a measurement function on a set of data -- returns targets and predicitons"""
     catalog_name = data_dir / "sample_input_catalog.fits"
     stamp_size = 24
     survey = get_surveys("LSST")
-    shifts = [[-0.3, 1.2]]
-    indexes = [[1]]
+    shifts = [
```
ismael-mendoza, line 23: I assume you added all these shifts to test it in  a lot more objects? :)

tests/test_measure.py
```diff
@@ -26,41 +47,54 @@ def get_meas_results(meas_function, cpus=1, measure_kwargs=None):
...
-        results["catalog"]["sep_measure"][0]["y_peak"].item(),
+    """Test sep dettection using single band and multiband with multiprocessing"""
+    target, results = get_meas_results(
+        [sep_singleband_measure, sep_multiband_measure], measure_kwargs=[{"sigma_noise": 1.5}]
     )
-    detected_centers = np.array([x_peak, y_peak])
-    dist = np.max(np.abs(detected_centers - target))
-    np.testing.assert_array_less(dist, 1.0)
+    detected_sources = {"sep_singleband_measure": 0, "sep_multiband_measure": 0}
+    for meas_function in detected_sources.keys():
```
ismael-mendoza, line 87: could you add a small comment explaning what this block that follows is doing exactly?

tests/test_measure.py
ismael-mendoza: How did you pick them btw?

ismael-mendoza (changes requested):
This looks great @atorchylo ! Thanks so much for implementing this :) 

I just had a couple of small suggestions/questions below which are mostly about adding comments/documentation. Please let me know if you have any questions about any of these. 

One more thing, could you go through the notebooks in the `notebooks` directory and do a quick scan to see if any of them need changing? Thank you again!

btk/measure.py
atorchylo: Hmm, I am not sure what would be the best approach here. 

This is how the original `sep_measure` worked on multiresolution input, so I just decided to copy this part into `sep_multiband_measure` function. It also implemented the same way in other measurement functions like `basic_measure`. 

Should we change this to iteration over all surveys?

btk/measure.py
atorchylo: In general `match_to_ctalog_sky` matches each pair of coordinates in `c1` to k-nearest neighbours in `c2`. In our case, we are only interested in the closest coordinate, so `k=1` 

- `d2d` - is a matrix with the shape `[len(c1), k]` that contains angular distances on the sky between each coordinate in `c1` and its neighbors.

- `d3d` - is the same as `d2d` but it computed 3d distance between objects if `z` was provided, otherwise it assumes that all coordinates are equally far away from us.

I added few comments and changed `d2d` to `distance2d` to make code more readable.

btk/measure.py
atorchylo: Sure! Added in the new commit.

btk/measure.py
atorchylo: Yes, exactly! `ra_coordinates` is the running "combined" list of coordinates. And  each time I get new detections, I make sure that they are no closer then `matching_threshold` before adding them to the "combined" list of coordinates.

tests/test_measure.py
atorchylo: I was verifying that `sep_multiband_measure` detected more sources than `sep_singleband_measure`. I also was trying to check that each prediction was close to the truth. Let me know if there is a better way of testing, or if you have any specific tests in mind.

tests/test_measure.py
atorchylo: Yeah... I just found these shifts in `test_metrics.py` and changed it to have one galaxy in a blend, so it would be easier to match sep detection to the ground truth.

btk/measure.py
ismael-mendoza: Thanks!

btk/measure.py
ismael-mendoza: Thanks @atorchylo I realized my original comment was a bit vague. I was thinking of something simpler, like allowing the user to specify explicitly which of the surveys they want to use in the case of MR. 

This is part of a broader issue I see with the MR functionality that is currently implemented, we can discuss this more sometime if you are interested but I think this is out of scope of this PR.

btk/measure.py
ismael-mendoza: hmm I don't think I see these changes? perhaps you forgot to push them?

btk/measure.py
ismael-mendoza: OK that makes sense to me, perhaps add a comment or two to this effect or in the docstring? I still find it a little hard to understand. Something like `# add new predictions for this band to catalog only if not close to existing centroid` ?

tests/test_measure.py
ismael-mendoza: great that makes sense to me, thank you!

btk/measure.py
ismael-mendoza: ah it was me looking at the wrong diff 😆 , this looks great now

btk/measure.py
ismael-mendoza: now that I looked at the correct diff, this is fine now

tests/test_measure.py
ismael-mendoza: now I get it, thanks! LGTM

ismael-mendoza (approved):

notebooks/01b-scarlet-measure.ipynb
```diff
@@ -397,9 +397,9 @@
  "metadata": {
   "celltoolbar": "Aucun(e)",
   "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
+   "display_name": "btk38",
    "language": "python",
-   "name": "python3"
+   "name": "btk38"
```
thuiop, line 402: We should avoid changing the metadata of the notebooks ; what did we want to do about that @ismael-mendoza ?

ismael-mendoza: oh yeah... I wonder why this hasn't happend before. Hmm let me see...

atorchylo: Oh I think that's because I used conda kernel when running this notebook. So, I switched default "python3" to "btk38"

I will fix it!

btk/measure.py
```diff
@@ -135,18 +137,120 @@ def basic_measure(
...
+    NOTE: If this function is used with the multiresolution feature,
+    measurements will be carried on the first survey.
+
+    Args:
+        batch (dict): Output of DrawBlendsGenerator object's `__next__` method.
+        idx (int): Index number of blend scene in the batch to preform
+            measurement on.
+        sigma_noise (float): Sigma threshold for detection against noise.
+        matching_threshold (float): Match centers of objects that are closer than
+            this threshold to a single prediction.
```
ismael-mendoza, line 167:
```suggestion
        matching_threshold (float): Match centers of objects that are closer than
            this threshold to a single prediction (in arseconds).
```

notebooks/01b-scarlet-measure.ipynb
ismael-mendoza: thanks!

Issue #340, opened 2022-07-27
Change the template of measure functions


atorchylo:
When I first encountered BTK measure function, I found it a little bit unintuitive because it required a very specific output, a dictionary with keys `catalog`, `deblended_images`, and `segmentation`, and I wasn't sure how to set the default parameters when passing a function to a `MeasureGenerator`. Currently, we can pass the parameter `measure_kwargs` to the `MeasureGenerator`, but it is unclear how this parameter works if we want to pass a list of different functions. The documentation says:
> Each dictionnary is passed one time to each function, meaning that each function which be ran as many times as there are different dictionnaries.

This could be very unintuitive and confusing for the users.

I propose that we use the format of PyTorch Image transforms, which uses classes to implement similar functionality. In this format, every measure function will be a class, with a `__call__`  method that implements what the function does, and `__init__`  that sets default parameters. So, in this format, we will rewrite the measure function into:
```
class Measure_smth()
   def __init__(self, default1=True, default2=True, default3=0.5):
      self.default1=default1
      self.default2=default2
      self.default3=default3

   def __call__(self, batch, idx):
      # do stuff
      return output_dict
```
So we can make an instance of this class and pass it to the `MeasureGenerator` as:
```
btk.measure.MeasureGenerator([Measure_smth(default1=False), Measure_smth(default1=True)], draw_blend_generator)
```

Also, we can write a small parent class for measure functions that would validate measure functions written by users, and give warnings if the user implemented the function correctly. In any case, this is just an idea, and I am happy to discuss it further!

ismael-mendoza:
I like this idea, what do you think @thuiop ?

thuiop:
This basically what we do for sampling functions anyway so I guess it would make sense ; however the idea behind the measure_kwargs was more about being able to run the same measure function with a whole array of parameters ; for instance for tuning a threshold. It was also meant to make the AUC metric work, although it has continuously been clunky. I agree that it is not super intuitive currently. I will try to think of a better way.

ismael-mendoza:
Implemented in #412

PR #348, opened 2022-08-19
Measure refactor

atorchylo:
This pull request will implement isssue #340 (currently, it's a work in progress). Specifically, it adds a base class `Measure` for all measurement functions. With this functionality, we can abandon `measure_kwargs` and switch to more intuitive object-oriented measure functions. We can use new measure objects as follows:

```
measures = [PeakLocalMax(),SepSingleband(), SepSingleband(use_mean=True), SepMultiband()]
btk.measure.MeasureGenerator(measures, draw_generator)
```

The current issue with this code is that it fails when a measure object returns `segmentation` or `deblended_images` in the multiresolution case. The current todo list for this pull request is as follows:

- [ ] Add a base class `MeasurementResult` that will do all of the type checking and standardize the output of the `Measure` function 
- [ ] Add unit tests and carefully test multiresolution feature
- [ ] Add docstrings and documentation 
- [ ] Fix tutorials & explain new functionality


ismael-mendoza:
TODO: Specify in the documentation that the `survey_names` for the `MeasuredExample`/`MeasureBatch` is about the output

atorchylo:
We want to refactor `target_meas` -> we have 4 options:
1) add `transforms` key argument to `MeasureGenrator` that will apply postprocessing to measurements
(+) more options for the user
(+) could be applied on both measured and original images
(-) makes `MeasureGeneroator` more complicated
(-) less intuitive for the user
(-) does basically the same thing as a `Measure`
2) do nothing / ask user to code up custom `Measure`, add an example in tutorials. 
(+) intuitive 
(+) requires no work on our side:)
(-) more work for the user
(-) not clear how to perform measurement on the original image 
3) do nothing / ask user to code up `Metric` that first measures and then runs a metric
(+) also intuitive
(-) user will not have access to measurements, only to final metrics 
4) standardize the interface between measure result and blend result and output the same thing
(+) clean
(-) not sure how to implement

We want:
1) Simple intuitive metrics
2) Customizability of metrics
3) Disentangle metrics from each other
4) Separate matching procedure
5) Add filtering step for certain galaxies (?)

Immediate To Do:
1) Set of metrics functions (simple inputs -> simple outputs)
2) Atomic metrics classes (measured_results, blend_results -> aggregated_outputs)
3) Take a look at TorchMetrics

ismael-mendoza:
- [ ] change `blend_list` to `catalog_list` and in measure change `catalog` to `catalog_list` 
- [ ] change matching to be ra/dec based? 
- [ ] add `match_fn` (Andrii) to Metrics
- [ ] Ismael finish Precision metric class, then split the rest
- [ ] Andrii give review of metrics_utils and match_fn

PR #366, opened 2022-09-09
allow metrics without matches

b-biswas:


b-biswas:
@ismael-mendoza, here I am just allowing the user to enter matches as `None` in the reconstruction metrics. 
In this case, it would assume a one-to-one mapping between the isolated and blended images.

btk/metrics.py
```diff
@@ -409,6 +409,13 @@ def reconstruction_metrics_blend(
                                           measurement function for each galaxy (both
                                           true and deblended images).
     """
+    if matches is None:
+
```
ismael-mendoza, line 413:
```suggestion
```

ismael-mendoza (approved):
Thanks @b-biswas for this PR - it looks good to me!

PR #373, opened 2022-09-30
Added fixed aperture flux target measure

### LINKED ISSUE #342: implement aperture photometry metric
@aguinot  suggested a very interesting metric based on aperture photometry which alleviate some important biases and should be interesting for photo-z folks as well. The procedure is as follows: 

input: noisy postage stamp with (simulated) galaxy blend 
output: aperture photometry flux for each object in the blend (if all were detected)

- Run deblending algorithm on image to get light profile of each detected object 
- For each detected object, take input noisy image and subtract all other objects from the image based on deblended (noiseless) profiles
- Measure the flux of that object using SEP aperture photometry (this is a function takes in an input image, a centroid, and the size of the aperture)

Finally, in order to do a fair comparison: Don't compare with true input flux from the catalog, rather repeat same procedure but subtracting the true noiseless light profiles of each object.

This procedure is especially important in the multi-band setting. 

Another interesting comparison, to see how much deblending would help, would be to just run this procedure with no deblending. Just run the SEP function on the input image without subtracting sources. 

**Note:** A similar metric for measuring shapes would also be good to implement, I think.

thuiop:
Closes #342 

New target measure in the metrics module, for measuring the flux using a fixed circular aperture with sep.


btk/metrics.py
```diff
@@ -112,6 +114,37 @@ def meas_ksb_ellipticity(image, additional_params):
...
+                                  and meas_band_num an integer indicating the band in which the
+                                  measurement is done.
+    """
+    meas_band_num = additional_params["meas_band_num"]
+    band = additional_params["survey"].available_filters[meas_band_num]
+    filt = additional_params["survey"].get_filter(band)
+    verbose = additional_params["verbose"]
+
+    bkg = sep.Background(image[meas_band_num])
+    catalog = sep.extract(image[meas_band_num], 1.5, err=bkg.globalrms)
```
aguinot, line 133: If you derive the background RMS from postage stamp I don't think the value will be reliable enough. I think it would be better to take the true value used in the simulation.

btk/metrics.py
```diff
@@ -112,6 +114,37 @@ def meas_ksb_ellipticity(image, additional_params):
...
+    if len(catalog) != 1 and verbose:
+        print(f"{len(catalog)} where detected when measuring flux.")
+
+    flux, fluxerr, flag = sep.sum_circle(
+        image[meas_band_num],
+        catalog["x"],
+        catalog["y"],
```
aguinot, line 140: I am not sure about this for the "reference" value. I think I would also take the true value from simulation for the object positions.

btk/metrics.py
```diff
@@ -112,6 +114,37 @@ def meas_ksb_ellipticity(image, additional_params):
...
+        filt.psf_fwhm.to_value("arcsec"),
+        err=bkg.globalrms,
```
aguinot, line 142: Same comment as for the detection.

btk/metrics.py
```diff
@@ -112,6 +114,37 @@ def meas_ksb_ellipticity(image, additional_params):
...
+    bkg = sep.Background(image[meas_band_num])
+    catalog = sep.extract(image[meas_band_num], 1.5, err=bkg.globalrms)
+    if len(catalog) != 1 and verbose:
+        print(f"{len(catalog)} where detected when measuring flux.")
+
+    flux, fluxerr, flag = sep.sum_circle(
+        image[meas_band_num],
+        catalog["x"],
+        catalog["y"],
+        filt.psf_fwhm.to_value("arcsec"),
```
aguinot, line 141: Maybe not require right now but it could be nice to be able to parametrize this.

btk/metrics.py
aguinot, line 133: If you are using small postage stamps I don't think you will be able to get a good estimate. I would rather take the true noise from the simulation.

btk/metrics.py
aguinot: Should be in pixel

ismael-mendoza:
1. fix pixel scale
2. compare apeture flux photometry (sextractor vs scarlet), scatter plot
3. documentation update
4. change labels of figures

thuiop:
I fixed the pixel scale issue and updated the doc ; I am merging this now and will open another PR for the plot (there are a few things I need to change, and I'll create a new notebook for reproducibility).

Issue #372, opened 2022-09-26
minimum number of galaxies


b-biswas:
Will it be a good feature to have a minimum number of galaxies in a blend along with the maximum number?
So we can keep a consistent galaxy density range useful for evaluating delbenders

ismael-mendoza:
Hi @b-biswas - sure I don't see a reason not to include this extra flag. It should be a simple addition to the default sampling function. Feel free to open a PR and implement this or let us know if you have any questions. Thanks!

PR #374, opened 2022-10-04
min_number argument in sampling func

b-biswas:
The goal of this PR is to address issue #372


ismael-mendoza:
Looks good to me ! Thanks

Issue #370, opened 2022-09-16
add rotation to sampling functions 


ismael-mendoza:
- reduce effect of having exact same object

ismael-mendoza:
**Thomas**: be careful about where to track the angle, perhaps add a `btk_rotation` parameter and apply it in the galaxy model level. Sampling function is the most logical place to sample and add it.

PR #376, opened 2022-10-07
Add rotation WIP

b-biswas:
This PR aims to make a random rotation to simulated galaxies. 
Addresses #370

### DIFF
--- a/btk/draw_blends.py
+++ b/btk/draw_blends.py
@@ -4,6 +4,7 @@ import os
 from abc import ABC, abstractmethod
 from collections.abc import Iterable
 from itertools import chain
+from math import pi
 
 import galsim
 import numpy as np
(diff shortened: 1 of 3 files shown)


aguinot (approved):
Everything looks good to me.
Have you check that you obtain the expected results?

One thing that I am thinking of is that the rotation will change the intrinsic ellipticity of galaxies. If this is something we are interested in or returned, we will have to propagate the rotation to the ellipticity. If we are only interested in the shear then it's fine.

ismael-mendoza:
Thanks @b-biswas - we discussed in the last hack hour and we decided that it would be best to include in the `DrawBlendsGenerator` as it's really a data augmentation feature and not something to be customized by the user. 

Thanks to your PR, I was able to very easily do this. So thank you! If you are curious here is the commit: https://github.com/LSSTDESC/BlendingToolKit/commit/1d862ee2481b482d723aba30d52dac2836f77ba7

PR #377, opened 2022-10-13
satisfying min_number condition

b-biswas:
PR #374 introduced some bugs in the number of galaxies that are being sampled. (the previous PR will give an error for DefaultSampling if `max_number` = 1
This PR is to fix the problem.


ismael-mendoza:
Thanks @b-biswas I incorporated your fix and some small changes in #382. I wanted to fix this sooner rather than later since it's a subtle bug

PR #378, opened 2022-10-14
Fixing SCARLET notebook and improving plotting

### LINKED ISSUE #364: revive scarlet notebook using `btkv1` tag in scarlet repo
need to look  at which was the last commit in pmelchior/scarlet that worked with the BTK scarlet notebook 

- [ ] check what happens if we only run it in single resolution mode

thuiop:
Closes #364.
The finally functionnal measure function for SCARLET is done and should stay that way thanks to the tag in the scarlet repo. It does not include multiresolution functionnality, as was discussed with Ismael and Axel.
Plotting the deblended images was also improved on the occasion, including colorbars and better overall presentation.

@pmelchior @fred3m may be interested in taking a look at this PR ; we manage to reproduce pretty good results with SCARLET in BTK (see following image).
resultsscarlet


ismael-mendoza:
- [ ] fixing the residual plotting 
- [ ] changing legend names in histogram to look nice
- [ ] reduce spacing between plots in reconstructions

thuiop:
Here is how the residuals look like currently.
residuals

ismael-mendoza:
That looks quite cool. The colorbar at the end of each row applies to the three images?

thuiop:
> That looks quite cool. The colorbar at the end of each row applies to the three images?
> 
> And could you clarify how you set the scale for the 'zoomed-in' residual? do you let matplotlib choose that one automatically?

Yes, the colorbar is the same for the 3 images ; it is obviously a different one for the zoomed-in one. The size of the zoom-in window is fixed ; the part that appears in it is currently fixed to be +- 12 pixels around the center of the galaxy. This is probably not ideal is you use a survey with a widely different pixel scale (e.g. HST) or deal with large galaxies, but we do not have these informations at this point in the code.

PR #386, opened 2022-10-18
fix SourceNotVisible exception

b-biswas:
The code crashes when using the catsim catalog because when the source is not visible in one of the bands, `render_single()` returns `None`. 
and `None` type has no attribute `None.array`


ismael-mendoza:
Thanks! Did this happen because you tried to use the bigger catalog? @b-biswas

PR #400, opened 2022-11-30
Check galaxy positions

### LINKED ISSUE #318: check that galaxies are shifted correctly in images
in particular the DefaultSampling function the max_shift = 0 does not work correctly

ismael-mendoza:
Closes #318 

- ensure that `maxshift=0` actually correct puts all galaxies in the center of the image 
- add test to check we can always detect bright galaxies in the correct location with SEP => galaxies are in the correct position in the image (as indicated by the `blend_list`)


ismael-mendoza:
Does this look ok to you @thuiop ?

ismael-mendoza:
I added a more careful check so now I definitely think it's correct and there is no bug, let me know however if you have any suggestions @thuiop

PR #412, opened 2023-03-03
Refactor v1

### LINKED ISSUE #167: Enable computation in multiple batches when saving results
It's unclear to me whether enabling computation in multiple batches when saving results would be at all useful. I was thinking that if users want to create a large # of blends, it might be useful to have >1 batch in order to avoid running out of memory (but I don't know what the upper limit on # of blends would be for a typical computer). 

Adding this feature would involve saving each batch separately and do some bookkeeping when saving and loading them which could be annoying. 

As it stands, I believe the codebase would just overwrite the previous batch if users tried to save more than one batch so at least we need to be clearer in the documentation on how saving works if we decide to not add this features. ( I don't think there is anything on the tutorials about saving?)

### LINKED ISSUE #171: Add SNR as a metric
It could be interesting to add the signal to noise ratio as a metric for galaxies ; however this is not completely trivial as we do not draw the isolated galaxies with the noise.

### LINKED ISSUE #280: Should we return lists or single objects?
Very difficult to handle a function that can return both a list and a single value.
You might want to chose one return type only (maybe in a future PR)

_Originally posted by @aboucaud in https://github.com/LSSTDESC/BlendingToolKit/pull/228#discussion_r830946784_

### LINKED ISSUE #395: add scarlet notebook back to CI
now that @thuiop fixed it and we have a tag in the scarlet repo

### LINKED ISSUE #396: remove `channels_last` option
I think it's ok for the codebase to follow a specific convention and to inform users of such convention so that they can modify the output themselves.

### LINKED ISSUE #419: augment data bug
it looks like augment data is rotating galaxies differently in different bands:
For example, here is a [plot](https://github.com/b-biswas/BlendingToolKit/blob/check-rotation/notebooks/rotation_bug.ipynb)  for the same field in 2 different bands. 
Also, `blend_list` stores rotation information only in the last band.

ismael-mendoza:
Copied over from #348 due to a mistake when making a release...

- Closes #396 
- Closes #171 
- Closes #167 
- Closes #280 
- Closes #395 
- Closes #419 

## Priority 

- [x] allow DeblendExample to use multiple bands (see #421)
- [x] test and add to tutorial `Scarlet` deblender

## ToDo 

- [x] fix tests (minimal)
- [x] fix pylint linting
- [x] linting keep: pylint, flake8, isort, pydocstyle, nbstripout, black
- [x] add minimal docs, delete old parts
- [x] delete or cleanup notebooks
- [x] fix README (minimal)
- [x] keep precommit?
- [x] need to add token to publish in pypi (`publish.yml`)


ismael-mendoza:
We want to refactor target_meas -> we have 4 options:

    add transforms key argument to MeasureGenrator that will apply postprocessing to measurements
    (+) more options for the user
    (+) could be applied on both measured and original images
    (-) makes MeasureGeneroator more complicated
    (-) less intuitive for the user
    (-) does basically the same thing as a Measure
    do nothing / ask user to code up custom Measure, add an example in tutorials.
    (+) intuitive
    (+) requires no work on our side:)
    (-) more work for the user
    (-) not clear how to perform measurement on the original image
    do nothing / ask user to code up Metric that first measures and then runs a metric
    (+) also intuitive
    (-) user will not have access to measurements, only to final metrics
    standardize the interface between measure result and blend result and output the same thing
    (+) clean
    (-) not sure how to implement

We want:

    Simple intuitive metrics
    Customizability of metrics
    Disentangle metrics from each other
    Separate matching procedure
    Add filtering step for certain galaxies (?)

Immediate To Do:

    Set of metrics functions (simple inputs -> simple outputs)
    Atomic metrics classes (measured_results, blend_results -> aggregated_outputs)
    Take a look at TorchMetrics

ismael-mendoza:
- [ ]  change blend_list to catalog_list and in measure change catalog to catalog_list
- [ ] change matching to be ra/dec based?
- [ ] add match_fn (Andrii) to Metrics
- [ ] Ismael finish Precision metric class, then split the rest
- [ ] Andrii give review of metrics_utils and match_fn

ismael-mendoza:
I spent some more time thinking about metrics yesterday and pushed some new classes of metrics. I'm not completely satisfied with the current metric structure >_< , but perhaps it's a good start. 

Here are some thoughts that we can discuss next time we meet @atorchylo 

- Is the metrics class doing too much? Should we separate the matching from computing the metrics? The difficulty is that some metrics like precision and recall use the matches explicitly to compute their corresponding quantity.
- I was thinking it would be nice if there was an option to use the metric without `BlendBatch` and `MeasureBatch` in addition to the __call__ function that takes these ones explicitly. However, I have difficulty atomitizing the metrics classes into separate simpler methods because of the matching. I also have trouble deciding what the inputs should be.
- Every metric has a `_get_data` and `all_data` that stores the data necessary to compute the metric, e.g. for precision this is a dictionary containing at least `tp` and `p`. The dictionary is a bit opaque in what it contains, should we expose it in the __str__ and __repr__ ? or in the documentation of the class? It might be useful for the user to know what intermediate data products are stored. 
- It could be useful to have a tutorial showing how to compute the metrics from `metrics_utils.py` rather than from `metrics.py`?
- Should the `compute` method always return a float if possible (i.e. except for the efficiency matrix)
- for some like the `iou` I was thinking it would be nice to compute segmentation before the metric (so we don't need an additional `seg_fn` or something like that). I was thinking of creating a separate `source_measure.py` that computes ellipticity, aperture photometry, and segmentation based on deblended images contained within `BlendBatch` and `MeasureBatch`.

ismael-mendoza:
@atorchylo just had a very useful meeting with @aguinot on feedback for our refactoring. Let's touch base before adding any new code, thanks.

ismael-mendoza:
@atorchylo here are the main takeaways from my conversation with Axel. We can discuss more on Tuesday but I wanted to give you a heads up: 

- Probably easiest to have a separate `match.py` and a corresponding `Matching` class, with functions implemented in `match.py` that output this class based on well-defined inputs. This should be thought as a step of the BTK pipeline, not a pre-processing step in the metrics. 
- The metrics can be separate classes as we have thought about to track internal state over batch computation. The metrics will take `Matching`, `BlendBatch` and/or `DeblendBatch`. This solves the problem of having to internally compute the matching for every metric.
- Consider a `metrics` folder containing `segmentation.py`, `reconstruction.py`, `detection.py` for each type of metric
- function that makes it easy to `btkize` user input, e.g. ensure user inputs catalos/images in a correct format and create `BlendBatch` / `DeblendBatch`. This means that we can just assume metrics take on this standarized input. 
- One important downside that Axel pointed out for our `BlendBatch` and `DeblendBatch` classes is that currently they require to specifically use the name of the survey you put in to access data. This makes it so that if we switch to another survey like HSC, the user would need to manually update the code in many places. Instead we should make this more like a "list" or in the single survey case "a single object list", which would make this much better. We can do fun things with the python class so that we can still access it with `.LSST` if we wan to keep that but it shouldn't be necessary.
- `ngmix` is a good resource for how they structure their code to be maintananble. In particular, note how they handle Observations and multiple Observations. (https://github.com/esheldon/ngmix/blob/51c488b0914a4499b9372875576a463e2697b29d/ngmix/metacal/convenience.py#L98)
- for plotting, rely as much as possible on existing plotting packages rather than rewriting from scrach which makes it hard to maintain. 
- for MR, specific enough that it's good to have a separate class with MR on its name that can do this
- for case when we want to use the truth catalog as a reference, just add `reference` keyword to deblender

atorchylo:
1. Create `BlendBatch` and `MultiresolutionBlendBatch`. This simplifies our simple deblenders, so we can remove the `survey_name` argument from our deblenders. 
2. Ask someone to cook up an example of a multi-resolution deblender and add it to `deblenders.py`
3. Add to the tutorials explanation about the output of the `Matching` class 
4. Standardize the input to `BlendBatch`, `DedeblendBatch`

atorchylo:
@ismael-mendoza I pushed two commits that implement a matching procedure for `DeblendBacth`. I decided to have a method `match` in the base class of `DeblendBatch` that will modify in-place the data that is stored in the deblended batch. It is not the cleanest piece of code, but I couldn't come up with anything better at this point. Please let me know if you have any suggestions.

ismael-mendoza:
Notes 05/17/23 

- Initial idea for BlendBatch as a list, better yet will be to create a separate "MRBlendBatch" class that actually cares about survey name and keeps track. But hard to disantangle as MR needs to keep all together in blend generator

ismael-mendoza:
05/19/23 

Current ToDo: 

- [ ] turn assertions in `blend_batch.py` into error messages 
- [ ] should we keep `survey_name` in `BlendBatch`? Could remove and just have `image_size` like in `DeblendBatch` / `DeblendExample`. 
- [ ]

ismael-mendoza:
05/31/23 

## TODO 

- [ ] Add some way of quickly checking how many sources are available/matched after `.match()` method. Perhaps just a wrapper around `.isnan()` along a certain axis or if not something clever.

ismael-mendoza:
## Notes - June 9, 2023

BTK future plans:

- **Potentially for ShearFest:** Testing shear bias with BTK / Realistic blend positions 
- another cool thing: catsim => realistic clustering => on top do SSI
- potential evolution of BTK: shear bias. 1) Testing step (e.g. check model bias, blending effects negligible); 2) DC2 create simulation as observing the sky, run entire pipeline and large fields; 3) Calibration (small fields) 
- having a setup that is oriented for ML methods to train NN. Create a tensorflow dataset (option). Really focus on that to really encourage people to use it.

ismael-mendoza:
07/04/23 

## Notes

- shouldn't deblended images be per band?

ismael-mendoza:
## Updates so far (July 12, 2023) 

- New `BlendBatch` class that organize output from DrawBlendGenerator, including catalog, psf, blend images, isolated images,... 
- New `Deblender` class that replaces `measure_functions` enforcing a specific structure for the output and input (now `BlendBatch` object)
- New `DeblendBatch` class that organizes output from `Deblender`, it also naturally allows for multiprocessing
- User has the option to do multiprocessing or not in `Deblender`, e.g. if using ML method they can perform operation in batches.
- added Scarlet as a new `Deblender` class
- added Scarlet dependencies as optional install (will add instructions for installing Scarlet directly, as no pypi)
- Separation between matching and calculating metrics. Metrics now assume that you have matched objects, catalogs first.
- New `match.py` module that implements metrics for performing 1-1 matches on pixel or sky coordinates.
- New `metrics` folder that includes detection, segmentation and reconstruction metrics. 
- New `measure.py` module that implements ellipticity measurement, SNR, blendedness, 'residual images'

ismael-mendoza:
## Idea for computing detection quantities as a function of SNR, etc.

- Create arrays of `detections` and `matches`, arrays of 1 or 0 for each object in true and predicted respectively. Each value in the array is 1 if the object was matched, 0 otherwise
- Since each entry in each array correspond to an object, we can assign it e.g. an SNR value
- If we want to compute a detection metric for a given SNR bin, just mask the corresponding array to only the values that correspond to the given e.g. SNR range.

ismael-mendoza:
## ToDo 

- [ ] fix tests (minimal)
- [ ] recreate large sample by picking a sub-region in catsim (rather than a random sample of all galaxies) 
- [ ] add SamplingFunction that returns random ~1' squares
- [ ] allow DeblendExample to use multiple bands (see #421)
- [ ] test and add to tutorial `Scarlet` deblender
- [ ] add minimal docs, delete old parts

PR #415, opened 2023-03-08
Make plots for the flux

thuiop:
This is the PR I have been working on, fixing some of the plots for the paper and making the one for the flux. Pre-commit has been failing for some unknown reason so this may need some lintering.


thuiop:
@ismael-mendoza this should be good now.

PR #443, opened 2023-08-18
Implements RandomSquareSampling

### LINKED ISSUE #423: new sampling function: subregions sampling
- [ ] recreate large sample by picking a sub-region in catsim (rather than a random sample of all galaxies) 
- [ ] add SamplingFunction that returns random ~1' squares

atorchylo:
Added a `RandomSquareSampling` sampling function, which allows user to get a subset of input catalog of a needed size. Unlike `DefaultSampling`, which selects galaxies randomly and then shifts them on the output image, `RandomSquareSampling` selects galaxies that are physically close to each other.

Closes #423


btk/sampling_functions.py
```diff
@@ -332,3 +332,70 @@ def __call__(self, table: Table):
...
+        ra = blend_table['ra']
+        dec = blend_table['dec']
+        ra[ra > 180] = ra[ra > 180] - 360
+        dec[dec > 180] = dec[dec > 180] - 360
+        ra_range = ra.max() - ra.min()
+        dec_range = dec.max() - dec.min()
+
+        # convert size to arc deg
+        size = self.stamp_size / 3600
+        assert size <= min(ra_range, dec_range), \
```
ismael-mendoza, line 388: could you change the assert to a `if` statement raising a value error?

btk/sampling_functions.py
```diff
@@ -332,3 +332,70 @@ def __call__(self, table: Table):
...
+        ra_center = np.random.uniform(ra.min() + size/2, ra.max() - size/2)
+        dec_center = np.random.uniform(dec.min() + size/2, dec.max() - size/2)
+        ra_min, ra_max = ra_center - size/2, ra_center + size/2
+        dec_min, dec_max = dec_center - size/2, dec_center + size/2
+        (indecies,) = np.where((ra > ra_min) & (ra < ra_max) & (dec > dec_min) & (dec < dec_max))
+
+        # if there are more galaxies than maximum, randomly select max number out of indecies
+        if len(indecies) > self.max_number:
+            indecies = np.random.choice(indecies, self.max_number)
+        return blend_table[indecies]
```
ismael-mendoza, line 401:
```suggestion
        if len(indices) > self.max_number:
            indecies = np.random.choice(indices, self.max_number)
        return blend_table[indices]
```

ismael-mendoza: I'm debating about doing this actually, since random sampling can takeaway random adjacent galaxies. MIght it be better to raise an error and tell users to either increase `max_n_sources` or make the `stamp_size` smaller?

ismael-mendoza (changes requested):
Sorry for the delay @atorchylo I have some small suggestions and a question

btk/sampling_functions.py
atorchylo: Yeah, you are probably right about this. I changed it to raise an error.

ismael-mendoza (approved):

PR #460, opened 2023-09-19
Implements Friends of Friends Sampling function

### LINKED ISSUE #16: Implement friends-of-friends algorithm to generate blends

atorchylo:
Closes #16 
 
Using `fast3tree` python package, I implemented the friends of friends sampling function. The algorithm works by precomputing groups for the given table, and then randomly sampling a group according to `min_number` and `max_number` requirements. The user has to provide `link_distance` to the algorithm, which determines the grouping of the galaxies. If the group exceeds the stamp size, the function raises an error. Here is the example code to quickly test:

```
import btk
import numpy as np
import matplotlib.pyplot as plt

catalog = btk.catalog.CatsimCatalog.from_file('data/input_catalog.fits')

sampling_function = btk.sampling_functions.FriendsOfFriendsSampling(
    min_number = 5, 
    max_number = 10, 
    link_distance=2.5,
    stamp_size=24
)
survey = btk.survey.get_surveys('LSST')
draw_generator = btk.draw_blends.CatsimGenerator(
    catalog,
    sampling_function,
    survey,
    batch_size=8,
    stamp_size=24,
    njobs=1,
    add_noise="background",
    seed=0, # use same seed here
)

blend_batch = next(draw_generator)
ii = np.random.randint(0, 8)

# plot the first blend in the batch, with the r-band
plt.imshow(blend_batch.blend_images[ii, 2, :, :], cmap="gray")

# plot centers
plt.scatter(blend_batch.catalog_list[ii]["x_peak"], 
            blend_batch.catalog_list[ii]["y_peak"], c="r", marker="x")
```
Sampled galaxies look like this:
Screen Shot 2023-09-19 at 12 59 24 PM
Screen Shot 2023-09-19 at 1 00 07 PM


btk/sampling_functions.py
```diff
@@ -5,6 +5,8 @@
 import astropy
 import numpy as np
 from astropy.table import Table
+from astropy.coordinates import SkyCoord
+from fast3tree import find_friends_of_friends
```
ismael-mendoza, line 9: need to add this new package to the `pyproject.toml` and `poetry.lock`, if you don't know how to do this let me know and I can add it

btk/sampling_functions.py
```diff
@@ -429,3 +431,105 @@ def __call__(self, table: Table):
...
         return blend_table
+
+
+class FriendsOfFriendsSampling(SamplingFunction):
+    """Randomly selects galaxies using Friends Of Friends clustering algorithm.
+
+    Friends of friends clustering algorithm assigns a group of object same index if one can
+    each member of the group is within link_distance of any other member in the group.
+    This sampling function explicitly uses the spatial information in the input catalog to
+    generate scenes of galaxies. However, blends might not always be returned as a result.
```
ismael-mendoza, line 442: Maybe the caveat should be that depending on the `link_distance` you might or might not get blends.

btk/sampling_functions.py
```diff
@@ -429,3 +431,105 @@ def __call__(self, table: Table):
...
+        points = SkyCoord(table['ra'], table['dec'], unit=('deg', 'deg')).cartesian.xyz.value.T
+        group_ids = find_friends_of_friends(points, self._arcsec2dist(self.link_distance), reassign_group_indices=False)
+        table["friends_of_friends_id"] = group_ids
+
+    def __call__(self, table: Table):
+        """Samples galaxies from input catalog to make scene.
+
+        We assume the input catalog has `ra` and `dec` in degrees, like CATSIM does.
+        """
+        # group galaxies by indecies
```
ismael-mendoza, line 493:
```suggestion
        # group galaxies by indices
```

btk/sampling_functions.py
```diff
@@ -429,3 +431,105 @@ def __call__(self, table: Table):
...
+        blend_table = table[cond]
+
+        # filter by number of galaxies
+        num_galaxies = blend_table.to_pandas()['friends_of_friends_id'].value_counts()
+        cond1 = num_galaxies >= self.min_number
+        cond2 = num_galaxies <= self.max_number
+        if (cond1 & cond2).sum() == 0:
+            raise ValueError(
+                f"No groups with number of galaxies in [{self.min_number}, {self.max_number}], found using a `link_distance` of {self.link_distance}"
+            )
```
ismael-mendoza, line 512: could you fix the formatting here?

btk/sampling_functions.py
```diff
@@ -429,3 +431,105 @@ def __call__(self, table: Table):
...
+        self.stamp_size = stamp_size
+        self.link_distance = link_distance
+        self.max_number = max_number
+        self.min_number = min_number
+        self.max_mag = max_mag
+        self.min_mag = min_mag
+        self.mag_name = mag_name
+
+    @staticmethod
+    def _arcsec2dist(sep, r=1.0):
```
ismael-mendoza, line 478: Why do you need the optional argumentn `r`? and what does it represent?

ismael-mendoza (changes requested):
Looks really good @atorchylo thanks for implementing this. Sorry it took me a while to get back to you. Please find some suggestions below

ismael-mendoza:
One more thing @atorchylo could you try reducing the link length even further and produce some examples to see if you only get very dramatic blends. Thanks!

btk/sampling_functions.py
atorchylo: Yeah, it would be great if you could help me with that. Not sure how to do it

btk/sampling_functions.py
atorchylo: Yeah, i agree. We should probably just add few words about that in the docstring

btk/sampling_functions.py
atorchylo: spelling indices is my ultimate enemy :)

btk/sampling_functions.py
```diff
@@ -429,3 +431,105 @@ def __call__(self, table: Table):
...
+                f"No groups with number of galaxies in [{self.min_number}, {self.max_number}], found using a `link_distance` of {self.link_distance}"
+            )
+        group_id = np.random.choice(num_galaxies[cond1 & cond2].index)
+        indices = blend_table['friends_of_friends_id'] == group_id
+
+        # check if galaxies fit in the window
+        ra = blend_table["ra"]
+        dec = blend_table["dec"]
+        if (ra[indices].max() - ra[indices].min()) * 3600 > self.stamp_size or \
+        (dec[indices].max() - dec[indices].min()) * 3600 > self.stamp_size:
```
ismael-mendoza, line 520: consider splitting into `cond1` and `cond2` so it's easiesr to read

btk/sampling_functions.py
```diff
@@ -429,3 +431,105 @@ def __call__(self, table: Table):
...
+
+    @staticmethod
+    def _arcsec2dist(sep, r=1.0):
+        """Converts arcsec separation to a cartesian distance"""
+        return np.sin(np.deg2rad(sep / 3600.0 / 2.0)) * 2.0 * r
+
+    def _precompute_friends_of_friends_index(self, table: Table):
+        """Computes galaxy groups using friends of friends algorithm on cartesian coordinates"""
+        points = SkyCoord(table['ra'], table['dec'], unit=('deg', 'deg')).cartesian.xyz.value.T
+        group_ids = find_friends_of_friends(points, self._arcsec2dist(self.link_distance), reassign_group_indices=False)
```
ismael-mendoza, line 485: actually could you explain why you conver the link_distance in this way?

btk/sampling_functions.py
atorchylo: so to convert the distance between the two objects on the sky to ecludian distance you generally need three numbers delta r = r2 - r1, delta phi = phi2 - phi1, delta theta = theta2 - theta1. but we don't know distance to the objects necessarily, so the assumption for calculating distances is that they are all at the same distance from the origin (same redshift). r = 1 is really just a normalization factor, which tells you how how ecludian distances scale with delta theta and delta phi.

atorchylo: setting r = const for all objects seems like a rough assumption at first, but that's what we mean by a blend. we say that two galaxies are blended if they are close in delta phi and delta theta, and it doesn't matter that they are at a different redshifts

ismael-mendoza: that makes sense!

btk/sampling_functions.py
atorchylo: firends of friends is working in ecludian space, so i need to convert from link distance, which is given by the angular separation to ecludian distance.

atorchylo:
Hey @ismael-mendoza. I implemented your suggestions and also produced some examples of extreme blends. Code does seem to work as desired! Below are examples of galaxies blends with min number of 3 and link distance of 1.
Screen Shot 2023-11-26 at 10 42 52 PM
Screen Shot 2023-11-26 at 10 43 58 PM
Screen Shot 2023-11-26 at 10 45 19 PM

PR #499, opened 2024-06-24
Fix ksb ellipticity centroids

ismael-mendoza:
- [x] add centroids to ksb measurement function
- [x] when plotting, remove condition that measured shear needs to be |.| < 1
- [x] should take mean of ksb ellipticity measurements not median
- [x] #502 (mistakenly, but PRs are basically independent so no problem)


ismael-mendoza:
will move code to new PR

PR #502, opened 2024-06-25
Stamp size clarification

ismael-mendoza:
I have made it so that `stamp_size` only needs to be passed into sampling functions as a mandatory argument


ismael-mendoza:
all tests are passing merging into the other branch (due to a mistake)

PR #538, opened 2025-07-24
Update citation in README

thuiop:
Updates the recommended citation to use the OJA one rather than arXiv


ismael-mendoza:
sorry I was testing some things, will open a new PR as it is easier

